PR #2, opened 2019-02-13
Feature/st 71 ci pipeline

aaron-moore-ST:


.travis.yml
```diff
@@ -0,0 +1,28 @@
+language: node_js
+node_js:
+- '8'
```
pgs-kzasada: Specifying only the major version is enough? Don't we need exact version, e.g. 8.15.0? Could we use the latest LTS - 10.15.1?

aaron-moore-ST: Yes, our understanding is that travis will use the latest 8.x.x that it has available in nvm (like package.json "version": "^8").
We do need to update this to 10, but we'd rather allow the latest minor version rather than pinning to 10.15.1.

PR #17, opened 2019-02-19
Feature/st 88 behavioural ci

aaron-moore-ST:


package.json
```diff
@@ -15,7 +15,9 @@
     "dev": "webpack --config webpack.dev.js",
     "prod": "webpack --config webpack.prod.js",
     "start": "webpack-dev-server --color --progress --open --hot --config webpack.dev.js",
-    "hot": "npm-run-all dev start"
+    "hot": "npm-run-all dev start",
+    "testframework": "git clone https://github.com/SecureTrading/js-payments-testing; cd js-payments-testing",
```
james-simpson-ST: I'm not a big fan of this but I can't think of a neater way to include the test framework in our repo and pulling it into travis unless anybody else can?

Perhaps if it was pushed into maven but then we'd have to have a maven config and all the rest of it which feels excessive. So unless there's any better ideas I'm happy to go with this solution, at least it keeps this out of the travis config so we only have to define it once.

PR #19, opened 2019-02-20
Feature/st 69 communication iframe view

pgs-gmartowski:


src/core/helpers/dom.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Method for creating DOM elements
+ * @param type Type of element which we are creating
+ * @param id ID of element
+ * @param text Optional text for element
+ */
+const createFormElement = (type: string, id: string, text?: string) => {
+  const element = document.createElement(type);
+  element.setAttribute('id', id);
+  element.setAttribute('class', id);
+  element.innerHTML = text ? text : '';
```
james-simpson-ST: My javascript security is a little rusty, what would prevent XSS within the text parameter being passed into this function, can we guarantee that this would not for example be provided by the customer and contain markup which would then be rendered?

src/core/listeners/input.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+
+/**
+ * Method for checking validity object and sending response to parent node
+ * @param fieldId ID of field which is validated
+ */
+const inputValueListener = (fieldId: string) => {
+  let input = <HTMLInputElement>document.getElementById(fieldId);
+  window.addEventListener(
+    'message',
```
james-simpson-ST: Depending how many different types of messages we end up with at the end of all the APM/Cardinal integrations this might need to be split up more but while it's this size it's fine, I think the per field one will only ever need to cope with the form submit but for the control frame this will likely have lots of different actions it needs to cope with like start payment (for applepay/cardinal),  tokenising payment, authorise payment etc. But as this event handler will have to deal with all of these we will have to keep this in mind

But this is more of a comment for the future

examples/example.html
```diff
@@ -5,46 +5,48 @@
...
+        <label for="example-1-phone">Phone</label>
+        <input
+          id="example-1-phone"
+          type="tel"
+          placeholder="+48713545689"
+          required=""
+          autocomplete="tel"
+        />
+      </div>
+      <div id="st-card-number" class="st-card-number"></div>
```
pgs-kzasada: It looks that these lines were not formatted by Prettier.

examples/example.scss
```diff
@@ -11,6 +11,12 @@
     color: red;
   }
 
+  .st-card-number,
+  .st-expiration-date,
+  .st-security-code {
+    width: 100%;
```
pgs-kzasada: By default, all fields should be 100% wide and <div> element always is 100% wide, so I wonder why this definition is needed here.

src/core/classes/validation/ExpireDate.class.ts
```diff
@@ -1,40 +1,9 @@
...
-  static MONTH_KEYBOARD_CODE_FIRST_CHAR = [48, 49];
-  static DATE_MAX_LENGTH = 5;
-  static DATE_SLASH_PLACE = 2;
-  static DATE_EMPTY = 0;
-
   constructor() {
     super();
   }
-
-  dateInputMask(element: HTMLInputElement) {
```
pgs-kzasada: This was moved to another place? Because we should have masking for the expiration date.

src/index.ts
```diff
@@ -1,27 +1,22 @@
...
   const st = new ST('thisissupersecretidforaclient12344321');
-  const payment = new Payment(['ApplePay', 'GooglePay'], ['133456', '546565']);
-  document.body.appendChild(createFormElement('form', 'st-form'));
-  document
-    .getElementById('st-form')
-    .appendChild(
-      createFormElement('h1', 'st-form__title', 'Secure Trading Example Form')
-    );
+  const payment = new Payment(
+    ['ApplePay', 'Visa Checkout'],
```
pgs-kzasada: In the future, this will come from configuration sent by the merchant.

examples/example.scss
pgs-kzasada: @pgs-gmartowski do we need this definition here?

examples/example.ts
```diff
@@ -1,45 +1,27 @@
...
-  element.setAttribute('class', id);
-  element.innerHTML = text ? text : '';
-  return element;
-};
 const elementStyles = {
   style: {
     base: {
+      backgroundColor: '#fff',
+      border: '.1rem solid #ced4da',
+      borderRadius: '.4rem',
```
pgs-kzasada: Please don't focus on styles until we have final designs.

PR #20, opened 2019-02-20
Feature/st 59 expiration date

pgs-gmartowski:


src/core/classes/validation/ExpireDate.class.ts
```diff
@@ -1,9 +1,45 @@
...
+
   constructor() {
     super();
   }
+
+  /**
+   * Method for masking expiration date in format MM/YY
+   * @param element
+   */
+  dateInputMask(element: HTMLInputElement) {
```
pgs-kzasada: Here you have a mask for input element, but in the template, you are using `<select>`. In my opinion we should use `<input>` with mask.

pgs-gmartowski: yes, we were talking about it with guys from ST and input is much better solution here

src/core/classes/validation/ExpirationDate.class.ts
```diff
@@ -0,0 +1,54 @@
...
+   * @param fieldInstance
+   * @param event
+   */
+  public static dateInputMask(
+    fieldInstance: HTMLInputElement,
+    event: KeyboardEvent
+  ) {
+    const length = fieldInstance.value.length;
+    if (length < ExpirationDate.DATE_MAX_LENGTH) {
+      if (!ExpirationDate.KEYBOARD_CODES_FOR_DIGITS.includes(event.keyCode)) {
```
aaron-moore-ST: Can this condition make use of Validation.isCharNumber instead since it's performing the same job

aaron-moore-ST: Can this condition make use of Validation.isCharNumber instead since it's performing the same job

pgs-gmartowski: yes, I moved it to Validation class.

PR #21, opened 2019-02-20
Feature/st 60 security code

pgs-gmartowski:


examples/example.scss
```diff
@@ -11,18 +15,35 @@
...
 
+  .row,
   .st-card-number,
   .st-expiration-date,
   .st-security-code {
     width: 100%;
+    margin: 1rem;
+  }
+
+  &__input {
```
pgs-kzasada: Please don't focus on styles for now.

src/core/classes/validation/SecurityCode.class.ts
```diff
@@ -1,8 +1,50 @@
 import Validation from './Validation.class';
 
+/**
+ * Definition of security code validation
+ */
 class SecurityCode extends Validation {
+  static AMEX_SECURITY_CODE_LENGTH = 4;
```
pgs-kzasada: These configurations should be stored in one place to avoid mistakes. Can we reuse `cardTree` introduced by Aaron?

src/core/classes/validation/SecurityCode.class.ts
```diff
@@ -1,8 +1,50 @@
...
+    this.securityCodeLength = 3;
+  }
+
+  /**
+   * Method for validating length based on html attributes max and minlength
+   * @param validity Validation object of form
+   */
+  static isLengthCorrect(validity: ValidityState) {
+    if (validity.tooShort) {
+      return 'Security code is too short';
```
pgs-kzasada: English texts should not be hardcoded. Maybe we can create a `Language` class which has a static field named `translations` which is an object with translation key and translation value. And here code will look like `return Language.translations.VALIDATION_ERROR_SECURITY_CODE_SHORT`. What do you think?

components/securityCode/templates/index.html
```diff
@@ -16,13 +16,15 @@
           id="security-code"
           name="security-code"
           class="security-code__input"
-          placeholder="XXXX"
+          placeholder="XXX"
+          maxlength="3"
```
pgs-kzasada: This is not a constant value - it depends on the type of the card.

src/core/classes/validation/ExpirationDate.class.ts
```diff
@@ -0,0 +1,49 @@
...
+   */
+  public static dateInputMask(
+    fieldInstance: HTMLInputElement,
+    event: KeyboardEvent
+  ) {
+    const length = fieldInstance.value.length;
+    if (length < ExpirationDate.DATE_MAX_LENGTH) {
+      if (!Validation.KEYCODES_DIGIT.includes(event.key)) {
+        event.preventDefault();
+      }
```
aaron-moore-ST: We can refactor:
```
if (!Validation.KEYCODES_DIGIT.includes(event.key)) {
  event.preventDefault();
}
```
into:
```
Validation.isCharNumber(event);
```

src/core/classes/Payment.class.ts
```diff
@@ -23,7 +24,11 @@ class Payment {
   public arePaymentsAvailable() {
     this._payments.map(item => {
       if (!this._availablePayments.includes(item)) {
-        console.error(`${item} payment type is not available !`);
+        console.error(
+          `${item} ${
+            Language.translations.VALIDATION_ERROR_PAYMENT_IS_NOT_AVAILABLE
```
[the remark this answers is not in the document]
aaron-moore-ST: @pgs-gmartowski what do you think about this? is there a practical solution using template strings?

src/core/classes/validation/Validation.class.ts
```diff
@@ -28,10 +28,10 @@ class Validation implements IValidation {
 
   /**
    * Method for preventing inserting non digits
-   * @param event Keypress event
+   * @param event - Keypress event
    */
-  static isCharNumber(event: KeyboardEvent) {
-    if (!Validation.KEYS_DIGIT.includes(event.key)) {
+  public static isCharNumber(event: KeyboardEvent) {
+    if (!Validation.KEYCODES_DIGIT.includes(event.key)) {
```
pgs-kzasada: Could you use RegExp here? `\d` should work here.

pgs-gmartowski: that's true, done.

PR #32, opened 2019-02-26
Feature/st 109 expiration date field ut

pgs-gmartowski:


test/core/classes/validation/ExpirationDate.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+  let element = document.createElement('input');
+  let elementWithError = document.createElement('input');
+  let elementWithExceededValue = document.createElement('input');
+  element.setAttribute('value', correctValue);
+  elementWithError.setAttribute('value', incorrectValue);
+  elementWithExceededValue.setAttribute('value', correctDataValue);
+  return { element, elementWithError, elementWithExceededValue };
+}
+
+function eventFixture(eventType: string, eventKeyValue: string) {
```
aaron-moore-ST: parameter eventType is declared but not used

test/core/classes/validation/ExpirationDate.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    // when
+    beforeEach(() => {
+      instance = new ExpirationDate();
+      expirationDateClass = ExpirationDate;
+      validationClass = Validation;
+    });
+    // then
+    it('should new object be an instance od ExpirationDate and Validation class', () => {
+      expect(instance).toBeInstanceOf(ExpirationDate);
+      expect(instance).toBeInstanceOf(Validation);
```
aaron-moore-ST: should these be expirationDateClass and validationClass respectively? Otherwise they have not been used from above

test/core/classes/validation/ExpirationDate.spec.ts
pgs-gmartowski: true, I forgot to pass it into function 😃

pgs-gmartowski: fixed

test/core/classes/validation/ExpirationDate.spec.ts
pgs-gmartowski: That's true. I think we could also get rid of  `expirationDateClass` and `validationClass`; that's two variables less.

pgs-gmartowski: fixed

test/core/classes/validation/ExpirationDate.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    });
+  });
+
+  // given
+  describe('Method dateInputMask', () => {
+    // when
+    let expectedValueWithSlash: string,
```
pgs-kzasada: Please use one const let declaration per variable and add rule to TSLint `"one-variable-per-declaration": true`.

pgs-gmartowski: ok, done.

PR #38, opened 2019-02-28
Feature/st 24 api request

aaron-moore-ST:


src/core/classes/StCodec.class.ts
```diff
@@ -0,0 +1,101 @@
...
+      Math.random()
+        .toString(36)
+        .substring(2, length)
+    );
+  }
+
+  /**
+   * Add the wrapper data to the request object
+   * @param requestData The data to be contained in this request
+   * @return A JSON object ready to be encoded
```
james-simpson-ST: This isn't JSON yet it's a JavaScript object ready to be encoded as JSON

src/core/classes/StCodec.class.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+    return JSON.stringify(this.buildRequestObject(requestObject));
+  }
+
+  /**
+   * Decode the Json body from the fetch response
+   * @Param responseObject The response object from the fetch promise
+   * @return A Promise that resolves the body content (or raise an error casing the fetch to be rejected)
+   */
+  public decode(responseObject: Response | {}) {
```
james-simpson-ST: The STPP response returns in the following format: 

`
{
    "requestreference": "W33-0rm0gcyx", 
    "version": "1.00", 
    "response": [
        {
            "transactionstartedtimestamp": "2019-02-21 10:06:35", 
            "livestatus": "0", 
            "issuer": "SecureTrading Test Issuer1", 
            "splitfinalnumber": "1", 
            "dccenabled": "0", 
            "settleduedate": "2019-02-21", 
            "errorcode": "0", 
            "baseamount": "100", 
            "tid": "27882788", 
            "merchantnumber": "00000000", 
            "securityresponsepostcode": "2", 
            "transactionreference": "33-9-80168", 
            "merchantname": "webservice UNICODE merchantname", 
            "paymenttypedescription": "VISA", 
            "orderreference": "AUTH_VISA_POST-PASS-JSON-JSON", 
            "accounttypedescription": "ECOM", 
            "acquirerresponsecode": "00", 
            "requesttypedescription": "AUTH", 
            "securityresponsesecuritycode": "2", 
            "currencyiso3a": "GBP", 
            "authcode": "TEST56", 
            "errormessage": "Ok", 
            "operatorname": "webservices@securetrading.com", 
            "merchantcountryiso2a": "GB", 
            "maskedpan": "411111######0211", 
            "securityresponseaddress": "2", 
            "issuercountryiso2a": "US", 
            "settlestatus": "0"
        }
    ], 
    "secrand": "ZKAVMskWQ"
}
`

So we probably want to simplify this for everywhere else we call the sendRequest so that it unpacks the "responses" checks theres only 1 response (as that's all this interface supports) and then always returns just the flat JS object rather than this nested one? Does that make sense?

Basically the reverse of what you've had to do when building the request

test/core/classes/StTransport.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+      st = new StTransport(DEFAULT_PARAMS);
+    });
+    afterEach(() => {
+      fetchMock.resetMocks();
+    });
+
+    it('should return promise if fetched successfully', async () => {
+      fetchMock.mockResponse('{"errorcode": 0}');
+      await expect(st.fetchTimeout('', {})).resolves.toEqual(
+        new Response('{"errorcode": 0}')
```
james-simpson-ST: Needs to reflect an actual ST response

src/core/classes/StCodec.class.ts
```diff
@@ -0,0 +1,101 @@
...
+    if ('json' in responseObject) {
+      return responseObject.json();
+    }
+    throw new Error(Language.translations.COMMUNICATION_ERROR_INVALID_RESPONSE);
```
james-simpson-ST: So this checks that it's valid JSON returned but doesn't check if the errorcode is zero, should this actually parse out the error code and raise out with the gateway's response in the case of an error (the gateway's response should already be in the correct locale)

src/core/classes/StCodec.class.ts
```diff
@@ -0,0 +1,101 @@
+import Language from './Language.class';
+
+interface IStRequest {
```
james-simpson-ST: Also needs to take in the locale that the page is being displayed in as this is likely something that won't be in the jwt because it's something that the merchant might give the choice to their customers to pick

src/core/classes/StCodec.class.ts
aaron-moore-ST: It's not what our previous one did but I think it's more consistent with the transport/codec model so that makes sense to me.

test/core/classes/StTransport.spec.ts
aaron-moore-ST: don't know that it's needed for transport (since it should just be passing it to decode) but we do need this in the codec tests

src/core/classes/StCodec.class.ts
aaron-moore-ST: It's not what our previous one did but I think it's more consistent with the transport/codec model so that makes sense to me.

src/core/classes/StCodec.class.ts
aaron-moore-ST: I had assumed that adding the locale to Language would happen at some point when we get to that task, do you want a TODO to remind us to get it once it exists?

james-simpson-ST: Ah yes you are right and no the task itself should be good enough, don't think we want to duplicate TODOs and Tasks like we have done elsewhere ;)

test/core/classes/StTransport.spec.ts
james-simpson-ST: Yes you are right, correct comment wrong test ;)

src/core/classes/StTransport.class.ts
```diff
@@ -0,0 +1,89 @@
...
+   * We probably want to update this to use an AbortControllor once this is standardised in the future
+   * @param url The URL to be passed to the fetch request
+   * @param options The options object to be passed to the fetch request
+   * @param timeout The time (ms) after which to time out
+   * @return A Promise that resolves to a fetch response or rejects with a timeout
+   */
+  public fetchTimeout(
+    url: string,
+    options: object,
+    timeout = StTransport.TIMEOUT
```
james-simpson-ST: I was speaking to Lee about this and he reminded me of something quite important that our other API libraries deal with  (e.g. the python API) now I don't know whether this is an issue or not or even if it's something configurable in a browser which might not allow for this level of control but basically theres a difference between a connect timeout and a response timeout. You've implemented the response timeout perfectly but theres no connect timeout.

Is there any way to set this and also to ensure that connections are retried until the connect timeout is reached or does the fetch Api handle this for us? Basically we should make this transport do everything we say merchants should do when writing their own api connection:
https://docs.securetrading.com/document/toolbox/configuring-your-own-library/

src/core/classes/StCodec.class.ts
james-simpson-ST: Has this been resolved?

src/core/classes/StTransport.class.ts
aaron-moore-ST: you mean retry a timed-out connection up to x times or until it's been tried for y length? No, fetch doesn't handle that for you, but I'd be wary of implementing it without the abort controller since it'd mean we'd be spinning up several connections without closing off the ones we're ignoring.

Of course we do plan to implement AbortController once it's finalised...

james-simpson-ST:
Once the tests are passing this looks good, @aaron-moore-ST  can you take a look into why they failed on the PR?

PR #39, opened 2019-03-01
Feature/st 120 code refactoring split utils and helpers into classes

pgs-gmartowski:


src/core/classes/Element.class.ts
```diff
@@ -1,39 +1,35 @@
-import { createFormElement } from '../helpers/dom';
-import { defaultIframeStyle, iframesEndpoints } from '../imports/iframe';
-
-const { cardNumber, expirationDate, securityCode } = iframesEndpoints;
+import ST from './ST.class';
```
pgs-kzasada: In my opinion, there is no need to add `class` suffix.

pgs-gmartowski: should I change it here or we will change it later ?

src/core/classes/validation/CardNumber.class.ts
```diff
@@ -1,45 +1,127 @@
...
+    });
   }
 
   /**
-   * Validate a card number
-   * @param cardNumber The card nuber to validate
+   * Listens to submit event from Form and validate card number field
+   * @param fieldId
+   */
+  private postMessageEventListener(fieldId: string) {
```
aaron-moore-ST: Does it make sense for postMessage to be part of validation, would it make more sense for it to be part of Element?

test/core/classes/validation/CardNumber.spec.ts
```diff
@@ -19,16 +19,15 @@ each([
...
-  const cn = new CardNumber();
-  cn.brand = brand;
-  expect(cn.getCardLogo()).toEqual(expected);
-});
+each([[null, cardsLogos.chip], [{ type: 'fred' }, cardsLogos.chip], [{ type: 'AMEX' }, cardsLogos.amex]]).test(
+  'CardNumber.getCardLogo',
+  (brand, expected) => {
+    // const cn = new CardNumber('card-number');
+    // cn.brand = brand;
+    // expect(cn.getCardLogo()).toEqual(expected);
```
aaron-moore-ST: We have effectively removed several tests here, are we sure we want this?

src/core/classes/validation/CardNumber.class.ts
pgs-gmartowski: I think if we have credit card number stored in local storage there is no such need to listen to postMessage event.

pgs-gmartowski: Done

test/core/classes/validation/CardNumber.spec.ts
pgs-gmartowski: That's because of problem with manipulation in DOM - tests currently crashes only with this error; we have to figure out how to deal with that, but for now we moved this problem to task 121; that's why I have to comment out some UT.

james-simpson-ST:
FYI Tests will need to be passing before this will let you perform the pill request

src/core/classes/validation/SecurityCode.class.ts
```diff
@@ -1,11 +1,90 @@
...
-  constructor() {
+  get securityCodeLength(): string {
+    return this._securityCodeLength;
+  }
+
+  set securityCodeLength(value: string) {
+    this._securityCodeLength = value;
+  }
+
+  private _fieldInstance: HTMLInputElement;
```
pgs-kzasada: Could be marked as 'readonly'

src/core/classes/validation/SecurityCode.class.ts
```diff
@@ -1,11 +1,90 @@
...
+  private _securityCodeLength: string;
+  private static DEFAULT_SECURITY_CODE_LENGTH = '3';
```
pgs-kzasada: Can it be a number, not a string? You should also define type.

src/core/classes/validation/SecurityCode.class.ts
```diff
@@ -1,11 +1,90 @@
...
+
+  constructor(fieldId: string) {
     super();
+    localStorage.setItem('securityCodeValidity', 'false');
+    this._fieldInstance = document.getElementById(fieldId) as HTMLInputElement;
+    this.securityCodeLength = SecurityCode.DEFAULT_SECURITY_CODE_LENGTH;
```
pgs-kzasada: DEFAULT_SECURITY_CODE_LENGTH is a private field, so you can use `this`

src/core/classes/validation/SecurityCode.class.ts
```diff
@@ -1,11 +1,90 @@
...
+    this._securityCodeLength = value;
+  }
+
+  private _fieldInstance: HTMLInputElement;
+  private _securityCodeLength: string;
+  private static DEFAULT_SECURITY_CODE_LENGTH = '3';
+
+  constructor(fieldId: string) {
     super();
+    localStorage.setItem('securityCodeValidity', 'false');
```
pgs-kzasada: In my opinion setting localStorage item and _fieldInstance should be in separate methods.

src/core/classes/validation/ExpirationDate.class.ts
```diff
@@ -8,44 +8,71 @@ class ExpirationDate extends Validation {
...
+    localStorage.setItem('expirationDateValidity', 'false');
+    this._fieldInstance = document.getElementById(fieldId) as HTMLInputElement;
+    this.inputValidation(fieldId);
+  }
+
+  /**
+   * Aggregates keypress and postMessage events listeners
+   * @param fieldId
+   */
+  public inputValidation(fieldId: string) {
```
pgs-kzasada: `fieldId` parameter is not used

src/core/classes/Element.class.ts
pgs-kzasada: We will change it later.

james-simpson-ST:
As this is including the styling I wanted to take a look at the components styled within the example page, however when I run `npm run hot` I get the example page but the component frames are not available on their ports, how should I run this to be able to view all the components within the example page?

src/core/classes/validation/SecurityCode.class.ts
```diff
@@ -1,11 +1,90 @@
...
+
+  /**
+   * Listens to postMessage event from Form
+   */
+  private postMessageEventListener() {
+    const cardNumber = localStorage.getItem('cardNumber');
+    const securityCodeLength = Number(localStorage.getItem('securityCodeLength'));
+    if (SecurityCode.setInputErrorMessage(this._fieldInstance, 'security-code-error')) {
+      if (SecurityCode.cardNumberSecurityCodeMatch(cardNumber, securityCodeLength)) {
+        localStorage.setItem('securityCode', this._fieldInstance.value);
```
james-simpson-ST: I have been looking into localStorage a bit more and have found a fundamental problem with it. 

localStorage and sessionStorage both store the data set on it on the users hard drive. This means that the pan and cvv would both be written out to disk (even if it is for a short period).

The pan is not allowed to be written out to disk unencrypted and the cvv can _never_ be written out to disk and can only ever appear in ram (this is a PCI requirement). 

Therefore I believe we're going to have to re-consider using the postMessage which would only ever communicate the information rather than storing the information. This isn't as nice a solution but if we move it off to be handled by another class might not be too bad. 

I'm not sure what you guys think of whether this wants to be a new task in order to change this approach?

pgs-gmartowski: I'll let know Kamil about that, after being more focused by last couple of days on Visa Checkout and Cardinal Commerce I'm not sure if he continued my solution with localStorage or create something else.
I think that's a good topic for Monday's review.

pgs-kzasada: @james-simpson-ST you are totally right. I'll refactor this part during working on creating the first payment.

james-simpson-ST: That's fine I'll mark this as approved to allow this to continue

pgs-kzasada:
All changes from this pull request are here https://github.com/SecureTrading/js-payments/pull/47.

PR #40, opened 2019-03-01
Feature/st 29 integrate cc library

pgs-gmartowski:


src/core/classes/CardinalCommerce.ts
```diff
@@ -0,0 +1,199 @@
...
+      this._triggerLookupRequest();
+    });
+  }
+
+  /**
+   * Perform a THREEDINIT with ST in order to generate the Cardinal songbird JWT
+   */
+  public threedeinitRequest() {
+    return this.sendRequest({
+      sitereference: 'live2',
```
pgs-kzasada: Shouldn't we use the test site reference?

src/core/classes/CardinalCommerce.ts
```diff
@@ -0,0 +1,184 @@
...
+    ERROR: 'ERROR',
+    FAILURE: 'FAILURE',
+    NOACTION: 'NOACTION',
+    SUCCESS: 'SUCCESS'
+  };
+
+  private _cardinalCommerceJWT: string;
+  private _cardinalPayload: any;
+  private _cart: string[] = [];
+  private _payload: IStRequest = {
```
pgs-kzasada: Don't we need here a `cachetoken` to be sent in the payload? According to https://docs.securetrading.com/document/api/security/3-d-secure/ (Step 1) it should be.

example/index.html
```diff
@@ -50,6 +50,11 @@ <h1 class="example-form__title">Secure Trading<span>Example Form</span></h1>
       <div class="example-form__group">
         <button class="example-form__button" type="submit">PAY</button>
       </div>
+      <input
```
pgs-kzasada: What is the reason that JWT is stored in a hidden input and not inside JavaScript code?

example/index.ts
```diff
@@ -4,25 +4,8 @@
...
-
-  securityCode.create('securityCode');
-  const securityCodeMounted = securityCode.mount('st-security-code-iframe');
-
-  expirationDate.create('expirationDate');
-  const expirationDateMounted = expirationDate.mount('st-expiration-date-iframe');
-
-  st.registerElements(
-    [cardNumberMounted, securityCodeMounted, expirationDateMounted],
-    ['st-card-number', 'st-security-code', 'st-expiration-date']
```
pgs-kzasada: `st-card-number` is an ID of a form inside an iframe or a div in the example page?

src/core/ST.ts
```diff
@@ -1,7 +1,18 @@
...
+import Element from './Element';
+
+// @ts-ignore
+const jwt: string = document.getElementById('JWTContainer').value;
+
 /***
  * Establishes connection with ST, defines client.
  */
-class ST {
+class ST extends StTransport {
```
pgs-kzasada: `ST` class should not extend `StTransport`. I don't see any reason why you did this.

src/core/ST.ts
```diff
@@ -1,7 +1,18 @@
+import CardinalCommerce from './classes/CardinalCommerce';
+import StTransport from './classes/StTransport.class';
+import Element from './Element';
+
+// @ts-ignore
+const jwt: string = document.getElementById('JWTContainer').value;
```
pgs-kzasada: As I wrote above, I don't know why you use here hidden input field.

src/core/ST.ts
```diff
@@ -10,8 +21,36 @@ class ST {
   public static expirationDateComponent = '/expiration-date.html';
   public static securityCodeComponent = '/security-code.html';
 
-  constructor() {
+  // @ts-ignore
+  constructor(style: object, payments: object[]) {
+    const gatewayUrl = ST.GATEWAY_URL;
+    super({ jwt, gatewayUrl });
```
pgs-kzasada: `gatewayUrl` is an optional parameter, so if it is the same as in StTransport, there is no need to send it once again.

src/core/ST.ts
```diff
@@ -39,12 +78,23 @@ class ST {
...
+
+  /**
+   * Creates div for error or info messages
+   * @private
+   */
+  private _createMessageContainer() {
+    const div = document.createElement('div');
+    const body = document.getElementsByTagName('body')[0];
+    div.setAttribute('id', 'st-messages');
+    body.appendChild(div);
```
pgs-kzasada: I believe instead of creating a `body` variable, you can simply do it like this: `document.body.appendChild(div)`. Please also think about a more descriptive name for `div` variable.

src/core/classes/CardinalCommerce.ts
```diff
@@ -0,0 +1,184 @@
...
+ * Defines integration with Cardinal Commerce and flow of transaction with this supplier.
+ * Configuration steps:
+ * 1.Cardinal.setup() + required Merchant JWT
+ * 2.Cardinal.on('payments.setupComplete)
+ * 3.Add BIN detection to PAN field
+ * 4.Perform cmpi_lookup request (defined in STtransport class)
+ * 5.Cardinal.continue + required payload from cmpi_lookup response
+ * 6.Cardinal.on('pauments.validated) - process auth or return failure
+ */
+class CardinalCommerce extends StTransport {
```
pgs-kzasada: I think it should not extend `StTransport` class.

src/core/classes/CardinalCommerce.ts
pgs-kzasada: According to James, we need `cachetoken` here.

src/core/classes/CardinalCommerce.ts
```diff
@@ -0,0 +1,184 @@
...
+    accounttypedescription: 'ECOM',
+    expirydate: '01/20',
+    pan: '4111111111111111',
+    requesttypedescription: 'THREEDQUERY',
+    sitereference: 'live2',
+    securitycode: '123',
+    termurl: 'http://something.com'
+  };
+  private _threedeinitRequestObject: IStRequest = {
+    sitereference: 'live2',
```
pgs-kzasada: `sitereference` should come from the configuration object passed when creating the instance of ST library.

src/core/classes/CardinalCommerce.ts
```diff
@@ -0,0 +1,184 @@
...
+  private _threedeinitRequest = () => this.sendRequest(this._threedeinitRequestObject);
+
+  private _authCallToST = (authRequest: any) => this.sendRequest(authRequest);
+
+  /**
+   * Add Cardinal Commerce Songbird.js library to merchants page.
+   * When library is loaded then it triggers configuration of Cardinal Commerce
+   * @private
+   */
+  private _insertSongbird() {
```
pgs-kzasada: It looks like a generic method for inserting scripts, same like for VisaCheckout and Apple Pay integrations. It should be in a separate class.

src/core/classes/StCodec.class.ts
```diff
@@ -7,6 +7,7 @@ interface IStRequest {
   accounttypedescription?: string;
   expirydate?: string;
   securitycode?: string;
+  termurl?: string; // TODO shouldn't be needed for CC request but this needs to wait for 153 release
```
pgs-kzasada: Please change gateway URL to `https://webservices.securetrading.net/jwt/` and I think this should not be needed any more.

src/core/classes/CardinalCommerce.ts
```diff
@@ -0,0 +1,184 @@
...
+    SUCCESS: 'SUCCESS'
+  };
+
+  private _cardinalCommerceJWT: string;
+  private _cardinalPayload: any;
+  private _cart: string[] = [];
+  private _payload: IStRequest = {
+    accounttypedescription: 'ECOM',
+    expirydate: '01/20',
+    pan: '4111111111111111',
```
pgs-kzasada: Here you should pass only `cachetoken`, not all credit card details.

src/core/imports/cardinalSettings.ts
```diff
@@ -0,0 +1,48 @@
+const applePayConfig = {
```
pgs-kzasada: What is the purpose of having this file?

test/core/classes/CardinalCommerce.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  // given
+  describe('Method _onSetup', () => {
+    // then
+    it('should be called once', () => {});
+  });
+
+  // given
+  describe('Method _retrieveValidationData', () => {
+    // then
+    it('should be called once', () => {});
```
pgs-kzasada: Empty tests :(

src/core/classes/CardinalCommerce.ts
pgs-gmartowski: that's the configuration which I'd received from Aaron - @james-simpson-ST should we really use 'live2' value ?

src/core/classes/CardinalCommerce.ts
pgs-gmartowski: ok, then can I use some mock data for `cachetoken` ? Implementation of requesting it is in the other branch

example/index.html
pgs-gmartowski: that was my thought for the first time, but in Cardinal Commerce documentation they suggest to store it in hidden input; what is the purpose ? No idea :/
Link to CC documentation: 
**Step 5: Initialize It**
https://cardinaldocs.atlassian.net/wiki/spaces/CC/pages/131806/Getting+Started

pgs-gmartowski: @james-simpson-ST according to CC, why should we store jwt in hidden input ?

example/index.ts
pgs-gmartowski: that should be the name of div in example page, question is if those ids should be hardcoded or not ? From my point of view it could be named whatever merchant wants (I was looking at Stripe Elements, that's why I implement it like this)

src/core/ST.ts
pgs-gmartowski: true, but I had to move `GATEWAY_URL` to import file cause I need this endpoint in several places.

src/core/ST.ts
pgs-gmartowski: that's Cardinal Commerce suggestion, but we can discuss it tommorow

src/core/ST.ts
pgs-gmartowski: I exported it to `cardinalSettings.ts`

src/core/ST.ts
pgs-gmartowski: it seems that it was part of my previous error handling, sorry for not delete it :(
error handling has been resolved in this branch -  feature/ST-33-notification-frame

src/core/classes/CardinalCommerce.ts
pgs-gmartowski: withdrawn, I created new instance of `StTransport` and then use method `sendRequest` from that class.

src/core/imports/cardinalSettings.ts
pgs-gmartowski: that tends to be file with all cc configuration properties, now there is a couple of settings eg. urls. Should I move it somewhere else ?

test/core/classes/CardinalCommerce.spec.ts
pgs-gmartowski: yes :/ actually I've got the same problem as in Visa Checkout - I've got a global object which is hard to mock

src/core/classes/CardinalCommerce.ts
pgs-gmartowski: if I change this one I should also change `interface IStRequest`. I'm not sure but if first request shouldn't have these three fields specified and then in the second request should be only cachetoken ?

src/core/classes/StCodec.class.ts
pgs-gmartowski: termurl parameter is probably still needed - I receive an error if I don't add it

src/core/classes/CardinalCommerce.ts
pgs-gmartowski: for now I have to send object like:
`accounttypedescription: 'ECOM',
    expirydate: '01/20',
    pan: '4111111111111111',
    requesttypedescription: 'THREEDQUERY',
    securitycode: '123',
    sitereference: 'live2',
    termurl: 'http://something.com',
`
If I delete pan, securitycode, expirydate and add cachetoken instead of them, requets will not succeed.
@james-simpson-ST should this request look like that ?

src/core/ST.ts
pgs-gmartowski: done

src/core/classes/CardinalCommerce.ts
pgs-gmartowski: done, I added it as second parameter in constructor

src/core/classes/CardinalCommerce.ts
pgs-gmartowski: yup, I've added class `DomMethods` and there I have all functions which operates on DOM.

src/core/classes/CardinalCommerce.ts
```diff
@@ -0,0 +1,185 @@
...
+   * Listens to submit event, send request to ST and handle response
+   * @private
+   */
+  private _triggerLookupRequest() {
+    window.addEventListener('submit', event => {
+      event.preventDefault();
+      this.stTransport.sendRequest(this._payload).then((response: any) => {
+        this._cardinalPayload = {
+          AcsUrl: response.acsurl,
+          Payload: response.pareq // TODO this should be threedresponse not pareq but the server needs updating
```
james-simpson-ST: This will change in a future release but ST will be able to update this class when this happens

src/core/classes/StCodec.class.ts
james-simpson-ST: Yes termurl is waiting for a future release ST can update this in the future when this is updated

src/core/classes/CardinalCommerce.ts
james-simpson-ST: Where do we want to do this - should it be done on ST-33 or as another task?

example/index.html
pgs-gmartowski: ok, so I think we can store it in constructor instead of input.
Done.

src/core/classes/StCodec.class.ts
pgs-gmartowski: ok, so I leave it as it is

example/index.ts
pgs-gmartowski: ok so we agreed that ids should be specified by merchant

pgs-gmartowski: done

src/core/ST.ts
pgs-gmartowski: changed - now its attached by merchant in constructor

src/core/classes/CardinalCommerce.ts
pgs-gmartowski: I think we should have another task for that like `Payment flow - Submit - Cachetoken - CC - ST`.

src/core/classes/CardinalCommerce.ts
pgs-gmartowski: ok

pgs-gmartowski:
@james-simpson-ST, @pgs-kzasada - I added some comments and implement changes, can you look at PR once again ?

PR #44, opened 2019-03-05
Feature/st 35 visa checkout

pgs-gmartowski:


examples/example.html
```diff
@@ -47,6 +47,7 @@ <h1>Secure Trading Example Form</h1>
       <button type="submit" id="st-form__submit" class="st-form__submit">
         Pay securely
       </button>
+      <script src="https://sandbox-assets.secure.checkout.visa.com/checkout-widget/resources/js/integration/v1/sdk.js"></script>
```
james-simpson-ST: Is this because of the TODO in _attachVisaSDK? In which case is there another task that will address this?

src/core/classes/ST.class.ts
```diff
@@ -0,0 +1,68 @@
...
+      'st-expiration-date-iframe'
+    );
+
+    RegisterElements(
+      [cardNumberMounted, securityCodeMounted, expirationDateMounted],
+      ['st-card-number', 'st-security-code', 'st-expiration-date']
+    );
+
+    if (this._getAPMConfig('VISA')) {
+      new VisaCheckout(this._getAPMConfig('VISA'));
```
james-simpson-ST: How are we expecting this to be instantiated? Will the merchant have to create a new VisaCheckout object or will they include it in their standard config when calling ST?

I'm trying to get a picture of the interface for merchants so that we can document this

src/core/classes/VisaCheckout.class.ts
```diff
@@ -0,0 +1,107 @@
...
+    this._setConfiguration();
+  }
+
+  /**
+   * Loads Visa Checkout configuration as soon as script is loaded and button attached to DOM
+   */
+  private _setConfiguration() {
+    VisaCheckout._attachVisaButton();
+    V.init(this._initConfiguration);
+    VisaCheckout._paymentStatusHandler(
```
james-simpson-ST: Do we have another task to take these responses and fire off the next stage i.e. for SUCCESS to process the AUTH and for the CANCEL/ERROR to display this to the customer?

src/core/classes/VisaCheckout.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  /**
+   * Init configuration (temporary with some test data).
+   * apikey and encryptionKey will authenticate merchant.
+   * Eventually in config, there'll be merchant credentials provided, now there are some test credentials.
+   */
+  private _initConfiguration = {
+    apikey: '' as string,
+    paymentRequest: {
+      currencyCode: 'USD' as string,
```
james-simpson-ST: How will we get this information from the merchant, in theory this will be part of the JWT which while we cannot validate this hasn't been tampered with we can use the content out of this for the VCO display purposes 

Does this want to be a new subtask as this will require the JWT sent in by the merchant to be decoded to get the amount and currency out of the payload

src/core/classes/VisaCheckout.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  set paymentError(value: object) {
+    this._paymentError = value;
+  }
+
+  constructor(config: any) {
+    const {
+      props: { apikey }
+    } = config;
+    this._initConfiguration.apikey = apikey;
```
james-simpson-ST: We want to allow merchants to pass in overrides for more than just the apikey where we override the standard values that we use by default. E.g. any of the extra "Payment Request" fields. Aaron might be able to help with this to define what we want to allow to be overridden

src/core/classes/VisaCheckout.ts
```diff
@@ -0,0 +1,130 @@
...
+   * @param event
+   * @private
+   */
+  private _paymentStatusHandler(event: string) {
+    V.on(event, (payment: object, error?: object) => {
+      this.paymentStatus = event;
+      this.paymentDetails = payment;
+      this.paymentError = error ? error : {};
+      // TODO: we have to decide how will show status message to customer, for now there is some silly alert.
+      alert(`Status of payment: ${event}`);
```
james-simpson-ST: In the case where this is a success we actually need to finish the payment by processing an AUTH to our payment gateway (this will want to be done the same way as Kamil does the Merchant Immediate payment) but should use the card number/expiry etc returned by this Visa checkout response instead of the values from the components on the screen. 

FYI the same will be true at the end of the ApplePay flow

src/core/classes/VisaCheckout.ts
```diff
@@ -0,0 +1,130 @@
...
+/**
+ *  Visa Checkout configuration class; sets up VC APM.
+ *  The only data which merchant have to provide is apikey, rest of the configuration is setting automatically.
+ */
+class VisaCheckout {
+  private static VISA_CHECKOUT_BUTTON_PROPS: any = {
+    alt: 'Visa Checkout',
+    className: 'v-button',
+    role: 'button',
+    src: 'https://sandbox.secure.checkout.visa.com/wallet-services-web/xo/button.png'
```
james-simpson-ST: At the moment this will only allow merchants to process transactions to the sandbox URL (which is okay for testing) but they will need to be able to use the live system when processing live transactions. We usually require the merchant to tell us whether or not they are live or not in order to determine which URL to use. As a result we need to add to the config a flag of "livestatus" using values 0 and 1 and change this and the SDK_ADDRESS based on this live status

src/core/classes/VisaCheckout.ts
```diff
@@ -0,0 +1,130 @@
+declare const V: any;
+
+/**
+ *  Visa Checkout configuration class; sets up VC APM.
+ *  The only data which merchant have to provide is apikey, rest of the configuration is setting automatically.
+ */
+class VisaCheckout {
```
james-simpson-ST: Something to consider when we have ApplePay in place is whether or not there is a common interface to these types of payments (commonly referred to as Wallets) basically they all add a button somewhere, then start their libraries, then handle callbacks at the end so perhaps we want a parent class to share anything that is common between these wallet payments?

src/core/ST.ts
```diff
@@ -10,8 +18,37 @@ class ST {
   public static expirationDateComponent = '/expiration-date.html';
   public static securityCodeComponent = '/security-code.html';
 
-  constructor() {
+  // @ts-ignore
```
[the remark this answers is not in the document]
pgs-gmartowski: it was workaround because of `super()` which has to be just before `constructor` definition, but now when we don't need to extend this class there is no need to include this.

src/core/classes/VisaCheckout.ts
pgs-gmartowski: I wondered how we will pass currency and price to library and now I know :) 
Yes, I think there should be separate task for `JWT decoder`.

src/core/classes/VisaCheckout.ts
pgs-gmartowski: Ok. I ask Aaron about this properties.

src/core/classes/VisaCheckout.ts
aaron-moore-ST: FYI, the apple pay flow will process the AUTH synchronously (we come back after to tell apple how it's done) wheras the AUTH from the VisaCheckout flow can be asynchronous since we don't have to tell them in the browser about the result.

src/core/classes/VisaCheckout.ts
aaron-moore-ST: Long term (if we move to a session-based flow) this might be able to come back from the session-init in order to save us requiring livestatus from the merchant.

For now do we want to default it to 1 (live) unless they specify in the config?

src/core/classes/VisaCheckout.ts
pgs-gmartowski: Ok, I think for now I'll prepare function which will pass all the necessary data from VC response to AUTH. Then we should discuss with Kamil  the mechanism of AUTH, to prevent duplicating code.

src/core/classes/VisaCheckout.ts
pgs-gmartowski: true, I'll add it shortly

pgs-gmartowski: done, I prepare livestatus parameter, which is configured by merchant on library init.

src/core/classes/VisaCheckout.ts
pgs-gmartowski: That's what I thought, I think we can aggregate common methods and props after finishing both AP and VC. New, parent class should IMO be called `APM.ts`.
@james-simpson-ST, @aaron-moore-ST , @pgs-kzasada what do you think about that ?

src/core/ST.ts
```diff
@@ -10,8 +16,33 @@ class ST {
...
+
+    expirationDate.create('expirationDate');
+    const expirationDateMounted = expirationDate.mount('st-expiration-date-iframe');
+
+    ST.registerElements(
+      [cardNumberMounted, securityCodeMounted, expirationDateMounted],
+      ['st-card-number', 'st-security-code', 'st-expiration-date']
+    );
+
+    if (this._getAPMConfig('VISA')) {
```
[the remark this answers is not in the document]
pgs-gmartowski: true, done.

src/core/classes/VisaCheckout.ts
aaron-moore-ST: Wallet.class.ts might be safer as AP and VC both fit that category and it's less refactoring to add APM to the top than Wallet in the middle if we add a different kind of APM in the future?

src/core/ST.ts
```diff
@@ -10,8 +16,33 @@ class ST {
...
+
+    const cardNumberMounted = cardNumber.mount('st-card-number-iframe');
+
+    securityCode.create('securityCode');
+    const securityCodeMounted = securityCode.mount('st-security-code-iframe');
+
+    expirationDate.create('expirationDate');
+    const expirationDateMounted = expirationDate.mount('st-expiration-date-iframe');
+
+    ST.registerElements(
```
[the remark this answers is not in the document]
pgs-gmartowski: you mean, that it should be possibility to hide all form fields in configuration process ?

examples/example.html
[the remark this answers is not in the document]
pgs-kzasada: I was sure that we resolve this issue yesterday. I can look at this, but probably not tomorrow.

src/core/classes/VisaCheckout.class.ts
```diff
@@ -0,0 +1,107 @@
+const V = (window as any).V;
+
+/**
+ *  Visa checkout configuration class
+ */
+class VisaCheckout {
+  private static VISA_CHECKOUT_BUTTON_PROPS: any = {
+    alt: 'Visa Checkout',
```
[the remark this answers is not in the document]
pgs-kzasada: Customization of Visa Checkout button by the merchant is not a scope of this sprint, so we should set some default values.

src/core/classes/VisaCheckout.ts
pgs-kzasada: @james-simpson-ST we have in mind to create one generic class to have a common interface, so it will be easier to add another payment method in the future. We will go deeper with Apple Pay integration and then refactor the code.

src/core/ST.ts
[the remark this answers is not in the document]
pgs-kzasada: Moreover, in my opinion, this string should be stored in some constant value which can be imported by other modules if needed.

examples/example.html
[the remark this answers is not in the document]
pgs-gmartowski: yes, that is resolved :)

pgs-gmartowski: I resolve this discussion - I've wrote method which attaches sdk via js.

test/core/classes/VisaCheckout.spec.ts
```diff
@@ -0,0 +1,48 @@
+import VisaCheckout from '../../../src/core/classes/VisaCheckout.class';
```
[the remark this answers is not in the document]
pgs-gmartowski: the only UT which are left, are those in which we have global object from sdk

src/core/classes/VisaCheckout.ts
pgs-gmartowski: I added task `JWT Decoder` - https://securetrading.atlassian.net/browse/STJS-140

src/core/ST.ts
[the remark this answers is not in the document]
pgs-gmartowski: @pgs-kzasada true, done.

src/core/ST.ts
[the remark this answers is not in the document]
james-simpson-ST: Basically some merchants might want to use our Visa Checkout or Apple Pay without accepting credit card payments, I think this will be something to look at once we have all the payment options in the library and we look at merchant configuration as a whole

src/core/classes/VisaCheckout.class.ts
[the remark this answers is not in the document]
james-simpson-ST: Okay then can we have a task in to add this in during a future sprint?

src/core/classes/StTransport.class.ts
```diff
@@ -26,11 +26,10 @@ export default class StTransport {
   public get codec() {
     return this._codec;
   }
-  public static GATEWAY_URL =
-    'https://webservices.securetrading.net/public/json/';
+  public static GATEWAY_URL = 'https://webservices.securetrading.net/public/json/';
```
pgs-kzasada: @james-simpson-ST should we use the new URL https://webservices.securetrading.net/jwt/ or this one below?

src/core/classes/VisaCheckout.ts
```diff
@@ -0,0 +1,150 @@
...
+/**
+ *  Visa Checkout configuration class; sets up Visa e-wallet
+ */
+class VisaCheckout {
+  private static PROD_BUTTON_URL: string = 'https://secure.checkout.visa.com/wallet-services-web/xo/button.png';
+  private static PROD_SDK: string =
+    'https://secure.checkout.visa.com/checkout-widget/resources/js/integration/v1/sdk.js';
+  private static DEV_BUTTON_URL: string = 'https://sandbox.secure.checkout.visa.com/wallet-services-web/xo/button.png';
+  private static DEV_SDK: string =
+    'https://sandbox-assets.secure.checkout.visa.com/checkout-widget/resources/js/integration/v1/sdk.js';
```
pgs-kzasada: These addresses should come from some configuration and no be stored inside the class.

src/core/classes/VisaCheckout.ts
pgs-kzasada: @james-simpson-ST should be call CACHETOKENIZE first with data returned by Visa Checkout or AUTH request should have all of this data (without cachetoken)?

test/core/classes/VisaCheckout.spec.ts
```diff
@@ -0,0 +1,96 @@
+import VisaCheckout from '../../../src/core/classes/VisaCheckout';
```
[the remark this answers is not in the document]
pgs-kzasada: I haven't look at it yet.

src/core/classes/VisaCheckout.class.ts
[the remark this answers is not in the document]
pgs-gmartowski: There is a task for this in backlog - https://securetrading.atlassian.net/browse/STJS-144

test/core/classes/VisaCheckout.spec.ts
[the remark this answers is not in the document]
pgs-gmartowski: Ok. As we mentioned on daily scrum we can move it to another task with UT.

src/core/classes/VisaCheckout.ts
pgs-gmartowski: moved to `imports/apms.ts` -> `VISA_CHECKOUT_URLS`

src/core/classes/StTransport.class.ts
pgs-gmartowski: I think it should be this new one -> `https://webservices.securetrading.net/jwt/`
@james-simpson-ST should we use this one ?

src/core/ST.ts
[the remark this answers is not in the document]
pgs-gmartowski: I think it's out of this tasks scope. In backlog I added task for that - https://securetrading.atlassian.net/browse/STJS-146.
It's not a big deal, IMO it's one parameter in configuration plus condition in registerElements function.

src/core/classes/VisaCheckout.ts
pgs-gmartowski: @pgs-kzasada, @james-simpson-ST that's a good question. Anyway there should be some info that payment details come from APM

src/core/classes/VisaCheckout.ts
pgs-gmartowski: I think I've found all of those properties - https://docs.securetrading.com/document/api/digital-wallets/visa-checkout/api-configuration/, but shouldn't we move implementation of them into task -  STJS-144 ?

src/core/classes/VisaCheckout.ts
james-simpson-ST: @pgs-kzasada @pgs-gmartowski That is a very good point, after you mentioned sending through the card details on the AUTH the other day I have been thinking about this more and perhaps it would be better discussing this in the daily but I think in the case where we are processing the AUTH on behalf of the merchant there is actually no need to tokenise the payment because they are never going to need to carry on with these details. 

However in the case of the stop payment (i.e. where we want to return to the merchants server so they can finish the payment) we want to tokenise the request so they don't have to see the original card details, in this case we will want to tokenize the details no matter which APM it is coming from because the merchant still can't see the card details in this case. 

Then if the merchant is completely PCI compliant we want to have the option of turning tokenisation off because they are then allowed to see the card details and therefore will want to be able to perform their own checks based on the card number e.g. Card BIN look ups

PR #53, opened 2019-03-08
Feature/st 34 apple pay

pgs-gmartowski:


src/core/classes/ApplePay.class.ts
```diff
@@ -0,0 +1,77 @@
...
+ * 3. Call canMakePaymentsWithActiveCard(merchantID) to check if there is at least one card in Wallet.
+ * 4. Display Apple Pay button.
+ * 5. Construct ApplePaySession with versionNumber and ApplePayPaymentRequest as arguments.
+ * 6. Call begin() method to display the payment sheet to the customer and initiate the merchant validation process.
+ * 7. In onvalidatemerchant handler catch object to pass to completeMerchantValidation
+ */
+class ApplePay {
+  private _applePayVersionNumber: number = 3;
+  private _merchantIdentifier: string;
+  private _paymentRequestData: any = {
```
pgs-kzasada: Remember that this is a mock and real configuration should come from the configuration object sent by the merchant when creating an ST object.

src/core/classes/ApplePay.class.ts
```diff
@@ -0,0 +1,77 @@
...
+   * Sets Apple Pay button and begins Apple Pay flow
+   */
+  public setUpApplePayProcess() {
+    if (this.checkApplePayAvailability()) {
+      this.checkApplePayWalletCardAvailability().then(
+        (canMakePayments: boolean) => {
+          if (canMakePayments) {
+            this.paymentSetup();
+          } else {
+            return 'Apple payment is not available';
```
pgs-kzasada: The message should be translated and located in Language class.

src/core/classes/ApplePay.class.ts
```diff
@@ -0,0 +1,77 @@
...
+   * Defines Apple Pay session details and begins payment flow.
+   */
+  public paymentSetup() {
+    const session = new ApplePaySession(
+      this._applePayVersionNumber,
+      this._paymentRequestData
+    );
+    session.begin();
+    session.onvalidatemerchant = (event: any) => {
+      console.log(event);
```
pgs-kzasada: Remove console.log()

src/components/notification-frame/NotificationFrame.ts
```diff
@@ -2,7 +2,7 @@ enum messageTypes {
   error = 'ERROR',
   info = 'INFO',
   success = 'SUCCESS',
-  warning = 'WARNING'
+  cancel = 'CANCEL'
```
pgs-kzasada: Revert to 'warning'

src/components/notification-frame/NotificationFrame.ts
```diff
@@ -71,8 +71,8 @@ export default class NotificationFrame {
    * @private
    */
   public setAttributeClass() {
-    if (this.notificationFrameElement) {
-      this.notificationFrameElement.setAttribute('class', NotificationFrame._getMessageClass(this._message.type));
+    if (this.notificationFrameElement && NotificationFrame._getMessageClass(this._message.type)) {
```
pgs-kzasada: You can store in variable a result of NotificationFrame._getMessageClass()

src/core/classes/ApplePay.class.ts
pgs-gmartowski: done, I've made some nasty alerts there just for now; next step will be  implementation for NotificationFrame

src/core/classes/ApplePay.class.ts
pgs-gmartowski: removed

src/components/notification-frame/NotificationFrame.ts
pgs-gmartowski: reverted

src/components/notification-frame/NotificationFrame.ts
pgs-gmartowski: done

src/core/classes/ApplePay.class.ts
pgs-gmartowski: true, now config object from `index.ts` is assigned here

src/core/classes/ApplePay.class.ts
```diff
@@ -0,0 +1,386 @@
...
+  private _onInit(buttonText: string, buttonStyle: string) {
+    if (this.ifApplePayIsAvailable()) {
+      this.setApplePayVersion();
+      this.setSupportedNetworks();
+      this.setAmountAndCurrency();
+      this.setApplePayButtonProps(buttonText, buttonStyle);
+      this.addApplePayButton();
+      this.applePayProcess();
+    } else {
+      this.setNotification(MessageBus.EVENTS_PUBLIC.NOTIFICATION_ERROR, 'Apple Pay is not available on your device');
```
pgs-gmartowski: @pgs-kzasada please take a look at url problem which we discussed today, that looks like its lack of the slash at the end of url.

src/core/classes/ApplePay.class.ts
```diff
@@ -0,0 +1,383 @@
...
+  }
+
+  /**
+   * Sets payment sheet interactions handlers: onpaymentmethodselected, onshippingmethodselected, onshippingcontactselected
+   */
+  public subscribeStatusHandlers() {
+    this.session.onpaymentmethodselected = (event: any) => {
+      const { paymentMethod } = event;
+      this.session.completePaymentMethodSelection({
+        newTotal: { label: this.paymentRequest.total.label, amount: this.paymentRequest.total.amount, type: 'final' } // what is type ??
```
pgs-gmartowski: @aaron-moore-ST I haven't found what is 'type' parameter and how it should be determined

src/core/classes/ApplePay.class.ts
```diff
@@ -0,0 +1,383 @@
...
+    };
+  }
+
+  /**
+   * Handles onpaymentauthorized event and completes payment
+   */
+  public onPaymentAuthorized() {
+    this.session.onpaymentauthorized = (event: any) => {
+      this.session.completePayment({ status: ApplePaySession.STATUS_SUCCESS, errors: [] });
+      this.setNotification(MessageBus.EVENTS_PUBLIC.NOTIFICATION_SUCCESS, 'Payment has been authorized');
```
pgs-gmartowski: @aaron-moore-ST where should `AUTH` request be called ? Is this a proper place ?

test/core/classes/ApplePay.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  };
+  const jwt =
+    'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJsaXZlMl9hdXRvand0IiwiaWF0IjoxNTUzMjcwODAwLCJwYXlsb2FkIjp7ImJhc2VhbW91bnQiOiIxMDAwIiwiY3VycmVuY3lpc28zYSI6IkdCUCIsInNpdGVyZWZlcmVuY2UiOiJsaXZlMiIsImFjY291bnR0eXBlZGVzY3JpcHRpb24iOiJFQ09NIn19.SGLwyTcqh6JGlrgzEabOLvCWRx_jeroYk67f_xSQpLM';
+  const instance = new ApplePay(config, jwt);
+
+  return {
+    config,
+    jwt,
+    instance
+  };
```
pgs-gmartowski: @aaron-moore-ST, @pgs-kzasada UT are in progress - I just wanna be sure if the implementation is correct and I haven't missed sth.

src/core/classes/ApplePay.class.ts
```diff
@@ -0,0 +1,403 @@
...
+  }
+
+  /**
+   * Make a server-to-server call to pass a payload to the Apple Pay validationURL endpoint.
+   * If successful, Apple Pay servers will return a merchant session object which will be used in response to completeMerchantValidation
+   */
+  public onValidateMerchantRequest() {
+    this.session.onvalidatemerchant = (event: any) => {
+      this.validateMerchantRequestData.walletvalidationurl = event.validationURL;
+      const stTransport = new StTransport({ jwt: this.jwt });
```
aaron-moore-ST: Once we add in the AUTH process (onPaymentAuthorized) that will also need a transport so we probably want to move this to this.stTransport.
I'm not sure how that will play with the arrow functions being called from apple's code so it might be worth making that change to ensure that this won't break the current flow.

src/core/classes/ApplePay.class.ts
```diff
@@ -0,0 +1,403 @@
...
+      .postMessage({ type, content }, (window as any).frames[Selectors.NOTIFICATION_FRAME_IFRAME]);
+  }
+
+  /**
+   * Begins Apple Pay payment flow.
+   */
+  public paymentProcess() {
+    this.session = this.getApplePaySessionObject();
+    // begins merchant validate process
+    this.session.begin();
```
aaron-moore-ST: It could be safer to register the events before calling `session.begin()`. The API doesn't guarantee that it will be asychronous, and if it is not this would likely cause events to be fired before the handlers are registered.

src/core/classes/ApplePay.class.ts
pgs-gmartowski: true, moved to `this.stTransportInstance`

src/core/classes/ApplePay.class.ts
pgs-gmartowski: true, changed

src/core/classes/ApplePay.class.ts
pgs-gmartowski: you mean that handlers like: 
` this.session.onpaymentauthorized`
or 
`this.session.onvalidatemerchant = (event: any) => {`
should not use ES6 arrow notation ?

src/core/classes/ApplePay.class.ts
aaron-moore-ST: https://developer.apple.com/documentation/apple_pay_on_the_web/applepaylineitemtype
However since our acquirers don't currently support the "pending" option I believe this will only be able to be "final" for the foreseeable future.

src/core/classes/ApplePay.class.ts
aaron-moore-ST: I don't know which "this" would be the context when the function is called, whether it would be function which registered the handler (ours) or the one which calls it (applepaysession). It's just something to keep an eye out for when manual testing.

pgs-gmartowski:
> Looks like there's a few blank unittests, I don't know how much work is left there? Other than that I'm happy

Yes, I struggle with mocking `window.frames ` object and if I resolve this problem I will add all of those test suites implementations.

src/core/classes/ApplePay.class.ts
pgs-gmartowski: Yes, it's possible that standard ES5 functions are used in ApplePay library not without purpose.

src/components/notification-frame/NotificationFrame.ts
```diff
@@ -129,8 +129,25 @@ export default class NotificationFrame extends Frame {
...
+      this.setAttributeClass();
+    });
+
+    this._messageBus.subscribe(MessageBus.EVENTS_PUBLIC.NOTIFICATION_ERROR, (data: string) => {
+      this._message = { type: messageTypes.error, content: data };
+      this.insertContent();
+      this.setAttributeClass();
+    });
+
+    this._messageBus.subscribe(MessageBus.EVENTS_PUBLIC.NOTIFICATION_WARNING, (data: string) => {
```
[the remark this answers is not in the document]
pgs-kzasada: I'll fix that

src/environments/environment.ts
```diff
@@ -1,7 +1,7 @@
 export const environment = {
   production: false,
   testEnvironment: false,
-  FRAME_URL: 'http://localhost:8080',
+  FRAME_URL: 'https://10.10.31.72:8443/',
```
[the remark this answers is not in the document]
pgs-kzasada: I'll fix that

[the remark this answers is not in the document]
pgs-kzasada: I've updated README.md with the description of how run project properly.

src/core/classes/ApplePay.class.ts
```diff
@@ -0,0 +1,393 @@
...
+
+  /**
+   * All object properties are required for WALLETVERIFY request call to ST.
+   */
+  public validateMerchantRequestData = {
+    requesttypedescription: 'WALLETVERIFY',
+    walletsource: 'APPLEPAY',
+    walletmerchantid: '',
+    walletvalidationurl: '',
+    walletrequestdomain: window.location.hostname
```
pgs-kzasada: `window.location.hostname` is not supported by Internet Explorer. I'll add a task to implement workaround.

PR #75, opened 2019-03-21
Feature/stjs 21 tokenisation of data for transfer

pgs-kzasada:
The first commit contains the code for Control Frame and communication mechanism between iframes. I based this mechanism on Pub-Sub pattern and Post Messages. This MessageBus should be initialized on the merchant side because I need some global space to store all subscriptions. Moreover, there is no way to call iframe from another iframe without touching the merchant side - `window.parent`.


pgs-kzasada:
In my second commit, you can find `Payment` class with `cacheTokenize` method which is responsible for sending CACHETOKENIZE request with card details retrieved from iframes. I'll implement error handling when I'll work on STJS-23, because I need NotificationFrame which should be merged to `develop`. In STJS-23 I'll also change mocked amount and currency to real data.

src/core/shared/MessageBus.ts
```diff
@@ -0,0 +1,84 @@
...
+    EXPIRATION_DATE_CHANGE: 'EXPIRATION_DATE_CHANGE',
+    SECURITY_CODE_CHANGE: 'SECURITY_CODE_CHANGE'
+  };
+
+  static EVENTS_PUBLIC = {
+    ST_SUBMIT: 'ST_SUBMIT'
+  };
+
+  constructor() {
+    this._domain = window.location.href;
```
james-simpson-ST: With regards to the pub-sub pattern this sounds fine however we can add some security in here. Basically we know that the components will _all_ be hosted by Secure Trading on the webservices.securetrading.net domain this means we can lock down the domain to be hardcoded to webservices.securetrading.net for communication of these change events.

This won't be necessary for all events however e.g. form submits from the merchant page would be able to be sent into the control frame from the merchants page, then its up to the control frame to perform the same onsubmit event it would do normally

Basically we can ask the parent frame for what other frames to communicate with but the domain for publishing and subscribing from should always be webservices if it is a message type that can contain sensitive data and the parent frame shouldn't be able to subscribe to any message event which may involve sensitive data e.g. CARD_NUMBER_CHANGE.

james-simpson-ST: Thinking about it it's probably okay to keep this as window.location.href (which should be webservices.securetrading.net within the frames that call this) OR should we hardcode to webservices.securetrading.net to be safer (but would need to be overridden by the tests which is more awkward) ?

src/core/shared/Payment.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export default class Payment {
+  private stTransport: StTransport;
+
+  constructor() {
+    this.stTransport = new StTransport({ jwt: '' });
+  }
+
+  cacheTokenize(cardDetails: any) {
+    let requestBody = {
```
james-simpson-ST: For this request you shouldn't need to send in the sitereference, baseamount or currency as these will be provided by the JWT passed in by the merchant when they instantiate the library.

src/core/shared/Payment.ts
```diff
@@ -0,0 +1,24 @@
+import StTransport from '../classes/StTransport.class';
+
+export default class Payment {
+  private stTransport: StTransport;
+
+  constructor() {
+    this.stTransport = new StTransport({ jwt: '' });
```
james-simpson-ST: This should be the JWT passed in by the merchant when they instantiate the library.

example/index.html
```diff
@@ -0,0 +1,62 @@
...
+      </div>
+
+      <div class="example-form__spacer"></div>
+
+      <div class="example-form__section">
+        <div id="st-card-number" class="example-form__group"></div>
+        <div id="st-expiration-date" class="example-form__group"></div>
+        <div id="st-security-code" class="example-form__group"></div>
+        <div class="example-form__spacer"></div>
+        <div id="st-control-frame" class="example-form__group"></div>
```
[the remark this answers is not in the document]
pgs-kzasada: > As this frame will always be needed and be "invisible" to the merchant, it would be nice if the merchant didn't even know they needed to add a placeholder, we could just add the control frame to the end of the body for the merchant instead

This control frame has not only the business logic but also button to submit the checkout form. Because of that, it should be visible.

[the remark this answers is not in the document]
pgs-kzasada: @james-simpson-ST is it possible to ask one of your merchants which option is better/easier to integrate for them? Both options are good in my opinion and without different point of view I think it wil be hard to decide. What do you think?

src/core/shared/Payment.ts
pgs-kzasada: OK, but first we need a method for decoding JWT, am I right?

src/core/shared/Payment.ts
pgs-kzasada: Yes, I remember. I'll work on that during STJS-23. Is it OK?

src/core/shared/MessageBus.ts
pgs-kzasada: I'll review this mechanism once again from security point of view.

src/core/shared/Payment.ts
james-simpson-ST: In this case we let the Server Side handle the decoding of the JWT and just need to pass it straight through unmodified so when you change the StTransport to have the JWT (below comment) you won't need to send this data through anymore

src/core/shared/Payment.ts
james-simpson-ST: Yes that makes sense

src/core/shared/Payment.ts
james-simpson-ST: We only need to decode the JWT in the front end when contacting 3rd parties that don't know how to decode our JWT (i.e. Visa Checkout, CC and ApplePay)

example/index.html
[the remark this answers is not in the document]
james-simpson-ST: I will try and get some opinions on this, for now lets carry on with the button in the frame and we can always change it later 👍

src/core/shared/Payment.ts
pgs-kzasada: Thank you for clarification. Now I got your idea.

src/core/Element.ts
```diff
@@ -74,6 +58,7 @@ class Element extends ST {
   public mount(fieldId: string) {
     const iframe = Element.createFormElement('iframe', fieldId);
     iframe.setAttribute('src', this._iframeSrc);
+    iframe.setAttribute('name', fieldId);
```
pgs-kzasada: I had to introduce a new attribute to get the reference to the frame by name instead of index inside MessageBuss class.

src/core/shared/MessageBus.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  constructor() {
+    this.registerMessageListener();
+  }
+
+  private registerMessageListener() {
+    window.addEventListener('message', (event: MessageEvent) => {
+      let messageBusEvent: MessageBusEvent;
+
+      if (event.origin === window.origin) {
```
pgs-kzasada: @james-simpson-ST I've simplified communication between frames. It was not a good idea to use parent page as a proxy for communication between frames. Now frames communicate between each other directly. The list of subscribers is stored in SessionStorage (instead of a variable in parent page). The list does not contain any sensitive data. Before processing a message I also check if `event.origin` is the same as the origin of an iframe.

src/core/shared/MessageBus.ts
```diff
@@ -0,0 +1,60 @@
...
+  publish(event: MessageBusEvent) {
+    let frameOrigin = window.origin;
+    let subscribersStore = window.sessionStorage.getItem(MessageBus.SUBSCRIBERS);
+
+    subscribersStore = JSON.parse(subscribersStore);
+
+    // @ts-ignore
+    subscribersStore[event.type].forEach((frame: string) => {
+      // @ts-ignore
+      window.parent.frames[frame].postMessage(event, frameOrigin);
```
pgs-kzasada: Messages are sent only between frames.

src/core/shared/MessageBus.ts
```diff
@@ -0,0 +1,60 @@
...
+    });
+  }
+
+  subscribe(eventType: string, callback: any) {
+    let subscribers;
+    let subscriber = window.name;
```
pgs-kzasada: `window.name` is the name of the iframe.

src/core/shared/MessageBus.ts
pgs-kzasada: @james-simpson-ST please look at my newest commit. I've implemented more secure communication between frames stored sensitive data. I think there is no backdoor now.

PR #82, opened 2019-03-28
Feature/st 23 merchant immediately payment

pgs-kzasada:
Before merge this to `develop` all TODOs should be resolved. This is still work-in-progress because I have to think about how to safely use MessageBus for communication between frames and parent. But the whole payment process works.


src/components/control-frame/ControlFrame.ts
```diff
@@ -28,55 +35,100 @@ export default class ControlFrame {
...
+    this.onLoad();
   }
 
-  private onClick() {
-    this._payment.cacheTokenize({
-      pan: this._formFields.cardNumber.value,
-      expirydate: this._formFields.expirationDate.value,
-      securitycode: this._formFields.securityCode.value
+  private initEventListeners() {
+    this._buttonElement.addEventListener('click', (event: Event) => {
```
james-simpson-ST: Does this mean currently only clicking the submit will start a payment? In which case do we want to also listen to customers hitting enter on input fields (both on the merchants page and within each component)?

pgs-kzasada: Yes, it means that only `submit` event triggers payment request. I don't think we should listen to customers hitting the Enter. Or we should?

james-simpson-ST: I think we should be listening to submit/click events on each component input field, the merchants page and the control frames button basically making it feel like a standard form making the iframes invisible to the user. 

Unless I'm missing something I can only see us listening to the click event on the button so this would not be behaving like a normal form would for customers

src/core/classes/CardinalCommerce.ts
```diff
@@ -164,17 +145,48 @@ class CardinalCommerce {
...
   }
-}
 
-export default CardinalCommerce;
+  private _authorizePayment(threeDResponse: string) {
+    // @TODO: use MessageBus
+    this._controlFrame.postMessage({ type: 'AUTH', data: threeDResponse }, this._controlFrameUrl);
+  }
+
+  // /**
```
pgs-gmartowski: is this still needed ?

pgs-kzasada: I'l look is is needed and if not I'll remove with task STJS-25.

pgs-kzasada:
> Before merge this to `develop` all TODOs should be resolved. This is still work-in-progress because I have to think about how to safely use MessageBus for communication between frames and parent. But the whole payment process works.

With the newest commit, MessageBus is used for (secure) communication between frames and parent. I have to look at `Window.URL` and `Window.URLSearchParams` as they are not supported by IE.

src/components/control-frame/ControlFrame.ts
pgs-kzasada: I suggest a solution as below:
1. Remove the "Pay" button from ControlFrame.
2. ControlFrame should listen on PostMessage 'ST_SUBMIT` from merchant page and when it received a message it will start the payment process.
3. Add the "Pay" button to the merchant page and trigger 'ST_SUBMIT' PostMessage as a callback on an event SUBMIT.

What do you think?

pgs-kzasada:
I've created a story STJS-159 with all missing unit tests.

src/components/control-frame/ControlFrame.ts
james-simpson-ST: This sounds good, but additionally 

4. Each FieldFrame will need to listen for SUBMIT event on themselves and fire the 'ST_SUBMIT' PostMessage

james-simpson-ST:
I am happy for this to be merged back to the develop for ST-25 and 159 to be worked on separately 

However you'll probably have to comment out the failing unit tests in order to get past the pull requirements

Also I'm not sure if you've seen my comment [here](https://github.com/SecureTrading/js-payments/pull/83#discussion_r271303711) but we might want a sparate task to resolve the URL issue on this, ST-108 and ST-144 independently because we've both used that on these different branches.

src/components/control-frame/ControlFrame.ts
james-simpson-ST: This has now been completed

PR #83, opened 2019-03-28
Feature/st 108 customization parameters

james-simpson-ST:


src/core/shared/Frame.ts
```diff
@@ -0,0 +1,37 @@
+import { Styler, AllowedStyles, Styles } from './Styler';
+
+export default class Frame {
+
+    constructor() {
+    }
```
pgs-gmartowski: empty constructor

src/core/shared/Frame.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    public onInit() {
+        this._applyStyles();
+    }
+    
+    public parseUrl() {
+        var parsedUrl = new URL(window.location.href);
```
pgs-gmartowski: we should avoid using `var`, in this case you can use `const`

src/core/shared/Frame.ts
```diff
@@ -0,0 +1,37 @@
...
+        const styles: Styles = {};
+        parsedUrl.searchParams.forEach(function (value, param) {
```
pgs-gmartowski: an arrow function can be used here - `parsedUrl.searchParams.forEach((value, param) => {`

src/core/shared/Styler.ts
```diff
@@ -0,0 +1,84 @@
...
+            templates.push(`${tag} { ${tagStyle} }`);
+        }
+        return templates.join(" ");
+    }
+
+    public inject(styles: Styles) {
+        let style = document.createElement('style');
+        style.innerHTML = this._getStyleString(styles);
+        document.head.appendChild(style);
+    }
```
pgs-gmartowski: I think this method should be in `DomMethods.class.ts`

james-simpson-ST: Done

src/core/shared/Frame.ts
james-simpson-ST: Done

src/core/shared/Frame.ts
james-simpson-ST: Done

src/core/shared/Frame.ts
james-simpson-ST: Removed

example/index.ts
```diff
@@ -16,7 +16,12 @@ import { ST } from '../src/stjs';
       controlFrame: 'st-control-frame'
     },
     'st-notification-frame',
-    {},
+    {'background-color-input': 'AliceBlue',
+     'color-input-error': '#721c24',
+     'line-height-input': '12px',
+     'font-size-input': '12px',
```
pgs-kzasada: With this solution, it is impossible to define different styles, e.g. font-size, for different fields. I think we should allow the merchant to style each element as he wants. Maybe for now we can leave it as it is, but we should have in mind that this is not so flexible.

src/components/control-frame/ControlFrame.ts
```diff
@@ -33,11 +35,37 @@ export default class ControlFrame {
...
-  private onInit() {
+  public onInit() {
+    super.onInit();
     this.initEventListeners();
     this.initSubscriptions();
   }
 
+  protected _getAllowedStyles () {
+    let allowed = super._getAllowedStyles();
+    let button = '#' + Selectors.CONTROL_FRAME_BUTTON_SELECTOR;
```
pgs-kzasada: You can use ES2015 Template Strings

src/core/Element.ts
```diff
@@ -64,9 +66,14 @@ export default class Element {
    * @param elementName Name of input which we want to create
    */
 
-  public create(elementName: string) {
+  public create(elementName: string, styles: Styles) {
     this._name = elementName;
-    this._iframeSrc = Element.getComponentAddress(elementName);
+    let address = Element.getComponentAddress(elementName);
+    let params = new URLSearchParams();
```
pgs-kzasada: URLSearchParams are not supported by Internet Explorer, so we have to think about some polyfill.

aaron-moore-ST: Any idea what the compatability is like for the URL interface (eg. `const parsedUrl = new URL('http://www.example.com/demonstration.html');`)? I believe it's older and should be more widely supported.

example/index.ts
james-simpson-ST: Added to Jira backlog on task ST-158

src/core/Element.ts
pgs-kzasada: `Window.URL` is also not supported by IE. I'll think about some workaround.

james-simpson-ST: @pgs-kzasada I have been looking into this and have been hitting a bigger problem, it appears as if our current webpack/babel config does not work with IE

The problem seems to be the package I've used for the currency formatting on a previous task (money-ts) where this is not being transpiled by babel into ES2015, as a result the code currently crashes out in IE before I can even get to the URL code to try out a polyfill for that. This is already an issue on the develop branch :( 

Do you think you will be able to investigate this or do you think we should have a task at the top of next sprint to sort out compatibility within IE?

src/components/control-frame/ControlFrame.ts
```diff
@@ -33,11 +35,37 @@ export default class ControlFrame {
...
+    let buttonHover = button + ':hover';
+    allowed = { ...allowed,
```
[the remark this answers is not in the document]
james-simpson-ST: I feel it is better and more readable to see what we have actually set if we are explicit here, adding in more logic here feels a bit unnecessary when it wants to behave as a whitelist of only allow the merchant to override these specific styles, perhaps if this grew any more we'd want to do so but for now I feel this is manageable

src/components/notification-frame/NotificationFrame.ts
```diff
@@ -48,13 +51,64 @@ export default class NotificationFrame {
...
   }
 
-  private static ELEMENT_ID: string = 'st-notification-frame';
+  protected _getAllowedStyles () {
+    let allowed = super._getAllowedStyles();
+    let notification = '#' + NotificationFrame.ELEMENT_ID;
+    let error = '.' + NotificationFrame.ELEMENT_CLASSES.error + notification;
+    let success = '.' + NotificationFrame.ELEMENT_CLASSES.success + notification;
+    let warning = '.' + NotificationFrame.ELEMENT_CLASSES.warning + notification;
+    let info = '.' + NotificationFrame.ELEMENT_CLASSES.info + notification;
```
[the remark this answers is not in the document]
james-simpson-ST: Done

src/components/control-frame/ControlFrame.ts
james-simpson-ST: Done

src/core/Element.ts
pgs-kzasada: @james-simpson-ST I've created a separate task for this: STJS-168.

PR #88, opened 2019-04-01
Feature/st 144 visa checkout customizations

james-simpson-ST:


src/core/classes/VisaCheckout.ts
```diff
@@ -45,31 +47,73 @@ class VisaCheckout {
...
+      config = { ...config,
+                ...settings 
+               };
+    }
+    return config;
+  }
+
+  public _setButtonSettings(settings: any, stJwt: StJwt) {
+    let config = {locale: stJwt.locale};
+    if (settings != undefined) {
```
pgs-gmartowski: I think we can avoid code duplication in this case by adding function which will be look more, less like:
`private setConfiguration = (config, settings) =>  settings ? {...config, ...settings} : ''  `

src/core/shared/Utils.ts
```diff
@@ -42,6 +42,22 @@ export default class Utils {
...
+   * @param params The parameters as key, value pairs with which to update the URL
+   * @return The new URL including the extra parameters
+   */
+  public static addUrlParams(url: string, params: any) {
+    if (params != undefined) {
+      let urlParams = new URLSearchParams();
+      for (let param in params) {
+        urlParams.set(param, params);
+      }
+      return url + "?" + urlParams.toString();
```
pgs-gmartowski: i'm not sure but it seems like we can use template strings here: 
`return `${url}?${urlParams.toString()}`;`

aaron-moore-ST: As mentioned in pr#83 can we use the URL interface? That would be a sturdier method to build URLs if we can.

james-simpson-ST: I tried to use URL but this requires the URL to contain the full domain and path which I didn't want to hard code the Domain. This is good enough for the usecase and I've updated to use template strings. The only case that does not work is if the URL already contains parameters which for now we do not have this case therefore going with YAGNI we'll add that when we need it

src/core/classes/VisaCheckout.ts
james-simpson-ST: Done

src/core/shared/Utils.ts
pgs-kzasada: `URLSearchParams` won't work in IE. I'll look for a workaround. Moreover, using `new URLSearchParams()` you can pass an object with params as an argument, so `for` loop is not needed.

src/core/shared/StJwt.ts
```diff
@@ -44,6 +44,18 @@ export class StJwt {
...
+    /**
+    * Convenience getter for returning the locale
+    * @return The locale contained inside the encoded payload or en_GB if undefined
+    */
+    public get locale() {
+        let locale = "en_GB";
+        if (this.payload.locale) {
+            locale = this.payload.locale;
+        }
+        return locale;
```
pgs-kzasada: Minor thing, but you can simplify to `return this.payload.locale || 'en_GB'`. Are you sure that `this.payload` will be always defined?

src/core/shared/Utils.ts
```diff
@@ -42,6 +42,22 @@ export default class Utils {
...
 
+  /**
+   * Helper function to return a URL string which has parameters added to it
+   * @param url The current URL as a string
+   * @param params The parameters as key, value pairs with which to update the URL
+   * @return The new URL including the extra parameters
+   */
+  public static addUrlParams(url: string, params: any) {
+    if (params != undefined) {
+      let urlParams = new URLSearchParams();
```
pgs-kzasada: If `params` is an object you can pass this object as an argument to `new URLSearchParams` and for loop won't be needed.

src/core/shared/StJwt.ts
james-simpson-ST: Done (yes this.payload should always exist at this point)

src/core/shared/Utils.ts
james-simpson-ST: Done

PR #94, opened 2019-04-03
feature/ST 157 https server

james-simpson-ST:
This is a working https server but I have not been able to test with an Apple device as I am not working from home today. I have got it to the point where there is no errors in chrome when calling ApplePay but that is as far as I can test

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "scripts": {
     "build": "webpack --config webpack.prod.js",
     "start": "webpack-dev-server --color --progress --config webpack.dev.js",
-    "example": "live-server ./dist --port=9090 --host=localhost --no-browser",
+    "example": "ws --rewrite '/* -> /dist/$1' --https",
     "prod": "npm run build",
     "test": "jest --watch",
     "coverage": "jest --coverage",
@@ -49,7 +49,7 @@
     "jest": "^24.5.0",
     "jest-each": "^24.0.0",
     "jest-fetch-mock": "^2.1.1",
-    "live-server": "^1.2.1",
+    "local-web-server": "^2.6.1",
     "mini-css-extract-plugin": "^0.5.0",
     "node-sass": "^4.11.0",
     "postcss-import": "^12.0.1",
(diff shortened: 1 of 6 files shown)


james-simpson-ST:
N.B. I will also be updating the test framework to use https on the other repo

package.json
```diff
@@ -11,7 +11,7 @@
   "scripts": {
     "build": "webpack --config webpack.prod.js",
     "start": "webpack-dev-server --color --progress --config webpack.dev.js",
-    "example": "live-server ./dist --port=9090 --host=localhost --no-browser",
+    "example": "ws --rewrite '/* -> /dist/$1' --https",
```
pgs-gmartowski: how should I start application now to have https available ? I start only `example` profile and has some npm errors

james-simpson-ST:
The build is failing for the same tests as the apple pay branch and so we can pull this into the apple pay branch in it's current state of not passing the build

webpack.dev.js
```diff
@@ -7,7 +7,8 @@ module.exports = merge(common, {
   devtool: 'inline-source-map',
   devServer: {
     contentBase: './dist',
-    port: 8080,
+    port: 8443,
```
pgs-gmartowski: why exactly this port number ?

package.json
james-simpson-ST: Can you send me the traceback

You should also be able to run npm start as it should be able to run using the dev server still

webpack.dev.js
james-simpson-ST: On linux 443 is a protected port which only root can start, 8443 is commonly used for testing https via a port that doesn't need root to start the server

pgs-gmartowski: ok, thx :)

package.json
pgs-gmartowski: that's weird, now it's working:/ maybe I've got some problems with npm (I'm switching branches all the time).

pgs-gmartowski:
@james-simpson-ST  approved

package.json
aaron-moore-ST: I have had some trouble with needing an npm install after switching branches if they have different packages being added, that could be what you ran into.

pgs-gmartowski: yup, I think that could be the reason

PR #118, opened 2019-04-17
ST-183 New card images

pgs-gmartowski:


pgs-gmartowski:
> I had only one comment about images compression. Grzegorz is checking now if we can lower the size of this file.

@pgs-kzasada, @aaron-moore-ST It looks that images are now a bit smaller and have same quality :)

PR #119, opened 2019-04-18
Feature/st 116 applepay endpoint overridden

pgs-gmartowski:


src/core/integrations/ApplePay.class.ts
```diff
@@ -47,18 +47,19 @@ class ApplePay {
...
   public buttonStyle: string;
   public merchantId: string;
+  public merchantSession: any;
   public messageBus: MessageBus;
   public paymentRequest: any;
   public placement: string;
   public session: any;
-  public sitereference: string; // possibly not needed
-  public sitesecurity: string;
+  public sitereference: string; // ???
```
pgs-kzasada: @pgs-gmartowski could you confirm with Aaron if it is needed?

src/core/integrations/ApplePay.class.ts
```diff
@@ -280,7 +281,22 @@ class ApplePay {
   public onPaymentAuthorized() {
     this.session.onpaymentauthorized = (event: any) => {
       this.session.completePayment({ status: ApplePaySession.STATUS_SUCCESS, errors: [] });
-      this.setNotification(NotificationType.Success, Language.translations.PAYMENT_AUTHORIZED);
+      this.stTransportInstance
```
pgs-kzasada: Please move all the logic to `Payment` class. In my opinion, sending AUTH request and handling responses should be in the `Payment` class.

src/core/integrations/ApplePay.class.ts
james-simpson-ST: This shouldn't be needed as it will already be provided inside the merchant generated JWT and so doesn't need to be included again

src/core/integrations/ApplePay.class.ts
james-simpson-ST: Agreed, one place for AUTH requests, step payment logic and response handing :)

src/core/integrations/ApplePay.class.ts
```diff
@@ -280,7 +281,22 @@ class ApplePay {
...
+          walletsource: this.validateMerchantRequestData.walletsource,
+          walletmerchantid: this.validateMerchantRequestData.walletmerchantid,
+          walletvalidationurl: this.validateMerchantRequestData.walletvalidationurl,
+          walletrequestdomain: this.validateMerchantRequestData.walletrequestdomain
+        })
+        .then(() => {
+          this.setNotification(NotificationType.Success, Language.translations.PAYMENT_AUTHORIZED);
+        })
+        .catch(() => {
+          this.setNotification(NotificationType.Error, Language.translations.PAYMENT_AUTHORIZED);
```
james-simpson-ST: Shouldn't this be PAYMENT_DECLINED? Or maybe the field errorMessage in the response object from the ST server. Or is this down to STJS-25 once the above refactoring into the Payment class has been completed?

src/core/integrations/ApplePayMock.class.ts
```diff
@@ -0,0 +1,122 @@
+import { environment } from '../../environments/environment';
+import { NotificationType } from '../models/NotificationEvent';
+import Language from '../shared/Language';
+import ApplePay from './ApplePay.class';
+import DomMethods from '../shared/DomMethods';
+
+/**
+ * Mocked version of Apple Pay setting test environment for Apple Pay automated tests.
+ */
+class ApplePayMock extends ApplePay {
```
james-simpson-ST: Why was ApplePay done as a sub class where VisaCheckout was changed within the VisaCheckout class. 

I personally prefer the ApplePay way which keeps the Mock functionality separate to the live flow however shouldn't they be done in the same way? Perhaps we can refactor VisaCheckout to be done this way?

src/core/integrations/ApplePay.class.ts
pgs-gmartowski: Ok. Removed.

src/core/integrations/ApplePay.class.ts
pgs-gmartowski: I changed it just like in the visa checkout - it's now using Payment class.

src/environments/environment.test.ts
```diff
@@ -15,5 +15,10 @@ export const environment = {
     DEV_BUTTON_URL: 'https://sandbox.secure.checkout.visa.com/wallet-services-web/xo/button.png',
     DEV_SDK: 'https://sandbox-assets.secure.checkout.visa.com/checkout-widget/resources/js/integration/v1/sdk.js',
     MOCK_DATA_URL: 'https://merchant.example.com:8443/visaPaymentStatus'
+  },
+  APPLE_PAY_URLS: {
+    BUTTON_IMAGE:
```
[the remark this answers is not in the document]
pgs-gmartowski: it has dedicated url according to visa documentation, but looks exactly the same

src/core/integrations/ApplePayMock.class.ts
pgs-gmartowski: so do I; the Apple Pay mock implementation is much better, I'll try to change it this week.

src/core/integrations/ApplePay.class.ts
pgs-gmartowski: we have some set of messages which are common for every type of payment, those messages are only informative for user and IMHO should be as short and simple as possible.
@pgs-kzasada we can also add messages DECLINED as James said.

src/core/integrations/ApplePayMock.class.ts
pgs-gmartowski: I added task for that - https://securetrading.atlassian.net/browse/STJS-188

src/environments/environment.test.ts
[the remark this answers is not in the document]
pgs-kzasada: If it is the same, and if not depends on the environment, it should not be here. It should be stored somewhere else. Here we should only data which can differ between environments.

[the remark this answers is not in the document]
pgs-gmartowski: moved to images.ts; I moved there also images of card logos

src/core/integrations/ApplePayMock.ts
```diff
@@ -0,0 +1,124 @@
...
+      this.setNotification(NotificationType.Error, `${errorcode}: ${errormessage}`);
+    }
+  }
+
+  /**
+   * Mocked AUTH process after this.session.completePayment()
+   * @private
+   */
+  private _mockedPaymentAuthorization() {
+    this.stTransportInstance
```
pgs-kzasada: Could you use `Payment.authorizePayment` method?

pgs-gmartowski: yes, I had to missed it somehow. Fixed.

PR #129, opened 2019-04-25
ST-172 Bin detection

pgs-gmartowski:


src/core/integrations/CardinalCommerce.ts
```diff
@@ -121,7 +121,15 @@ export default class CardinalCommerce {
     const messageBusEvent: MessageBusEvent = {
       type: MessageBus.EVENTS_PUBLIC.LOAD_CARDINAL
     };
-    this._messageBus.publishFromParent(messageBusEvent, Selectors.CONTROL_FRAME_IFRAME);
+
+    this._messageBus.subscribe(MessageBus.EVENTS.CHANGE_CARD_NUMBER, (data: FormFieldState) => {
```
pgs-kzasada: Yes, CardinalCommerce class should listen on change of card number and should call `bin.process`

src/core/integrations/CardinalCommerce.ts
```diff
@@ -121,7 +121,15 @@ export default class CardinalCommerce {
...
+      const { validity, value } = data;
+
+      if (validity) {
```
pgs-kzasada: CardinalCommerce class should not be responsible for checking if the card is valid or not. This is the responsibility of CardNuber card.

src/core/integrations/CardinalCommerce.ts
```diff
@@ -121,7 +121,15 @@ export default class CardinalCommerce {
...
+        Cardinal.trigger('bin.process', value);
+        this._messageBus.publishFromParent(messageBusEvent, Selectors.CONTROL_FRAME_IFRAME);
```
pgs-kzasada: Is it needed to inform ControlFrame about something? What is inside `messageBusEvent`?

src/core/integrations/CardinalCommerce.ts
pgs-gmartowski: true, removed

src/core/integrations/CardinalCommerce.ts
pgs-gmartowski: no, not really, I deleted it. As we discussed, I added new event BIN_PROCESS to check if Cardinal.trigger('bin.process') could be launched.

src/components/card-number/CardNumber.ts
```diff
@@ -49,13 +49,20 @@ export default class CardNumber extends FormField {
...
+
   private sendState() {
     let formFieldState: FormFieldState = this.getState();
     let messageBusEvent: MessageBusEvent = {
       type: MessageBus.EVENTS.CHANGE_CARD_NUMBER,
       data: formFieldState
     };
 
+    const binProcessEvent: MessageBusEvent = {
+      type: MessageBus.EVENTS.BIN_PROCESS,
```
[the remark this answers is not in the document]
pgs-gmartowski: true, done.

src/core/classes/StCodec.class.ts
```diff
@@ -95,6 +95,7 @@ class StCodec {
       throw new Error(Language.translations.COMMUNICATION_ERROR_INVALID_RESPONSE);
     }
     const responseContent = responseData.response[0];
+    console.log(responseData);
```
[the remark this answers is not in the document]
pgs-gmartowski: deleted

src/core/integrations/CardinalCommerce.ts
```diff
@@ -118,10 +118,9 @@ export default class CardinalCommerce {
    * This includes a failed JWT authentication.
    */
   private _onCardinalSetupComplete() {
-    const messageBusEvent: MessageBusEvent = {
-      type: MessageBus.EVENTS_PUBLIC.LOAD_CARDINAL
-    };
-    this._messageBus.publishFromParent(messageBusEvent, Selectors.CONTROL_FRAME_IFRAME);
```
pgs-kzasada: Why you removed this method?

pgs-gmartowski: oops, sorry my bad. Fixed.

src/components/card-number/CardNumber.ts
```diff
@@ -49,13 +49,20 @@ export default class CardNumber extends FormField {
...
+  private static getCardNumberForBinProcess = (cardNumber: string) => cardNumber.slice(0, 6);
+
   private sendState() {
     let formFieldState: FormFieldState = this.getState();
     let messageBusEvent: MessageBusEvent = {
       type: MessageBus.EVENTS.CHANGE_CARD_NUMBER,
       data: formFieldState
     };
 
+    const binProcessEvent: MessageBusEvent = {
```
james-simpson-ST: Should we not have some modified unittests to ensure this and CardinalCommerce have these calls?

We've been told by Cardinal that this is a critical part of the 3DS 2.0 implementation so we need to make sure this happens (even though it doesn't look like it's doing anything from a users perspective) so I think unittests are the best place for this.

pgs-gmartowski: yes, I agree, I'm trying now to write UT for these methods but it doesn't look to be simple (Cardinal Commerce and CardNumber class use MessageBus class instance to communicate).

pgs-gmartowski: but I'll do my best 😃 
@pgs-kzasada can you take a look at this ? Mocking functions it's not my strong point.

PR #140, opened 2019-04-29
[WIP] Feature/st 141 lock submit so multiple requests cannot be triggered

pgs-gmartowski:


james-simpson-ST:
I have been looking into this and from a visual point of view this is fine when the button locks but I have the following issues:

- If the payment fails the form needs to unlock again e.g. if the AUTH request declines
- Once the AUTH request is successful it'd be nice if we got rid of the Please wait and perhaps loaded in a success message like Payment complete
- I have noticed that if the button is clicked then the form/button is locked we also need to do this if any of the inputs have the enter key hit with this I think the below change should take place:

Instead of this being triggered inside the Form class it needs to move into the Payment class. Then it needs to use the messageBus to send a public message to the parent that will trigger the preloader. The reason for the messageBus is that now we have some requests (for Cardinal) that happen inside the control frame and some that happen (e.g. for ApplePay/VisaCheckout) that happen in the parent frame. If we use the message bus we can guarentee it gets fired by both. (We should also check that while we're processing a VisaCheckout the form is locked too)

Let me know if more explanation is needed

pgs-gmartowski:
> I have been looking into this and from a visual point of view this is fine when the button locks but I have the following issues:
> 
> * If the payment fails the form needs to unlock again e.g. if the AUTH request declines
> * Once the AUTH request is successful it'd be nice if we got rid of the Please wait and perhaps loaded in a success message like Payment complete
> * I have noticed that if the button is clicked then the form/button is locked we also need to do this if any of the inputs have the enter key hit with this I think the below change should take place:
> 
> Instead of this being triggered inside the Form class it needs to move into the Payment class. Then it needs to use the messageBus to send a public message to the parent that will trigger the preloader. The reason for the messageBus is that now we have some requests (for Cardinal) that happen inside the control frame and some that happen (e.g. for ApplePay/VisaCheckout) that happen in the parent frame. If we use the message bus we can guarentee it gets fired by both. (We should also check that while we're processing a VisaCheckout the form is locked too)
> 
> Let me know if more explanation is needed

Good point ! I'll check it once again and improve those things.

james-simpson-ST:
I haven't checked the code yet but this is looking much better from a usability point of view.

However I just tried to perform a case where we get an error on the Authorisation (basically the transaction is declined) and in this case I was expecting an error message to be displayed and the form to "unlock" to allow the customer to try an alternative card. 

Can you take a look into why this didn't work
You can test this case with our test card number 4242424242424242

pgs-gmartowski:
I close this PR - all changes were moved to ST-26

PR #141, opened 2019-04-29
ST-28 Step payment initiation

pgs-gmartowski:


src/core/integrations/VisaCheckout.ts
```diff
@@ -171,6 +191,49 @@ class VisaCheckout {
     }
   }
 
+  /**
+   * Starts AUTH request
+   */
+  protected _authorizePayment() {
```
james-simpson-ST: Could _authorizePayment and _cacheTokenizePayment be refactored to just call _processPayment which takes a function as paramenter e.g. _processPayment(this.payment.authorizePayment)?

src/core/integrations/ApplePay.ts
```diff
@@ -424,6 +427,53 @@ class ApplePay {
       });
     }
   }
+
+  /**
+   * Starts AUTH request
+   * @private
+   */
+  private _paymentAuthorize() {
```
james-simpson-ST: Could _paymentAuthorize and _cacheTokenizePayment be refactored to just call _processPayment which takes a function as paramenter e.g. _processPayment(this.payment.authorizePayment)?

src/core/integrations/CardinalCommerce.ts
```diff
@@ -183,6 +193,9 @@ export class CardinalCommerce {
       data,
       type: MessageBus.EVENTS_PUBLIC.AUTH
     };
+    if (this.step) {
+      messageBusEvent.type = MessageBus.EVENTS_PUBLIC.CACHETOKENISE;
```
james-simpson-ST: Does ControlFrame know how to handle this event? Can't see it in the diff?

pgs-gmartowski: good point ! I forgot about that.

src/core/integrations/ApplePay.ts
pgs-gmartowski: yes, I think it's possible, I take a look how AUTH and CACHETOKENISE requests look like and change it

src/core/integrations/CardinalCommerce.ts
pgs-gmartowski: done

src/core/integrations/VisaCheckout.ts
pgs-gmartowski: done, I had to refactor couple of functions but it looks like I haven't destroyed anything :)

src/core/integrations/ApplePay.ts
pgs-gmartowski: done

src/core/integrations/VisaCheckout.ts
```diff
@@ -114,6 +117,23 @@ class VisaCheckout {
     this._initConfiguration.settings = this.setConfiguration({ locale: stJwt.locale }, settings);
   }
 
+  /**
+   * Adds query string to src visa button image to customize it
+   * @param properties
+   */
+  public customizeVisaButton(properties: any) {
```
aaron-moore-ST: Did we get a URL shim for IE working? If so then it would be more robust & extensible to use that here rather than using string concatenation.

james-simpson-ST: I think we're currently working under the assumption that it will work and so yes it should be updated to use URL rather than string concatenation :)

pgs-gmartowski: ok, I changed the implementation and it uses now the URL object

PR #144, opened 2019-05-01
Feature/st 192 deploy to webservices

james-simpson-ST:
Following updates have been made
- Changes needed for Cardinal following updated server implementation
- Updated PROD URL to include path to where files will be hosted
- Updated domain checking in message bus to only use origin not origin+path
- Expose library as variable ST
- Fixed behavioural tests add commented the chrome one back in


test/core/shared/Utils.spec.ts
```diff
@@ -38,8 +38,8 @@ describe('timeoutPromise', () => {
     let after = before;
     await Utils.timeoutPromise(timeout).catch(e => (after = Date.now()));
     // toBeCloseTo is intended to check N significant figures of floats
-    // but by using -1 we can check it's within 5s of the set value
-    expect(after - before).toBeCloseTo(timeout, -1);
+    // but by using -2 we can check it's within 5s of the set value
```
aaron-moore-ST: This comment is no longer true, the -n checks that it's accurate to 1x10<sup>n</sup> around the first argument so this update will test that it's within 50s either side (that's a much greater margin, are we happy with that?)

test/core/shared/Utils.spec.ts
```diff
@@ -89,7 +89,7 @@ describe('retryPromise', () => {
     async (timeout, attempts, error) => {
       const promissory = jest.fn(() => Utils.timeoutPromise(timeout, error));
       await expect(Utils.retryPromise(promissory)).rejects.toThrow(error);
-      expect(promissory.mock.calls.length).toBeCloseTo(attempts, -1);
+      expect(promissory.mock.calls.length).toBeLessThan(attempts + 1);
```
aaron-moore-ST: Do we also want a greater than as well, since we also want to prove that it does actually retry (rather than only calling promissory once)?

webpack.common.js
```diff
@@ -17,6 +17,8 @@ module.exports = {
   output: {
     filename: '[name].bundle.js',
     path: path.join(__dirname, 'dist'),
+    library: '[name]',
+    libraryTarget: 'var',
```
aaron-moore-ST: Do we want to update index.html and example.ts in the repository to include stjs.bundle.js and call stjs.ST(stuff) rather than still having stjs bundled into example.bundle.js? This would catch it earlier if we broke this behaviour.

test/core/shared/Utils.spec.ts
james-simpson-ST: Aren't we talking milliseconds in JavaScript? so we're still 0.05s which I think is more than accurate enough

webpack.common.js
james-simpson-ST: Ideally yes but I feel this is a larger task that needs more consideration (as we'll have to change the environment.test.ts to allow for this so I'll add to backlog

test/core/shared/Utils.spec.ts
james-simpson-ST: I've looked at this a bit more and I think the test was actually doing the wrong thing entirely, can you take a look at my new version and make sure it actually tests what we think it does

aaron-moore-ST: Yes, looks good. Approval stands

PR #146, opened 2019-05-02
Feature/st fix ts warnings

james-simpson-ST:


src/components/animated-card/animated-card.ts
```diff
@@ -1,4 +1,6 @@
 import './animated-card.scss';
 import AnimatedCard from './AnimatedCard';
 
-AnimatedCard.ifCardExists() && new AnimatedCard();
+(() => {
```
aaron-moore-ST: one wonders if this should be an event listener for DOMContentReady

james-simpson-ST: @pgs-gmartowski do you know if this is needed or not?

pgs-gmartowski: I'm not sure but webpack wraps it in function, so IIFE can be redundant here.

PR #158, opened 2019-05-09
ST-196 Data attributes for notification

pgs-gmartowski:


pgs-gmartowski:
Some small changes for QA purposes - it helps Marek testing notifications.

PR #159, opened 2019-05-09
ST-198 Apple pay mock issues

pgs-gmartowski:


pgs-gmartowski:
After discussion with Marek we had to get rid of some notifications which spoils automated test.

PR #167, opened 2019-05-14
Bugfix/st manual testing

james-simpson-ST:
The following includes fixes required when applying the current working code to our development server ready for deployment

This was required when providing the code to our UAT testers


src/core/shared/Frame.ts
```diff
@@ -1,24 +1,30 @@
...
+    this._messageBus = new MessageBus(this._params.origin);
     this.applyStyles();
   }
 
   public parseUrl() {
     const parsedUrl = new URL(window.location.href);
     const styles: IStyles = {};
     const params: IParams = {};
-    const allowedParams = ['locale'];
+    const allowedParams = ['locale', 'jwt', 'origin'];
```
aaron-moore-ST: It might benefit security to only allow jwt on the controlframe (I'm pretty sure it should be the only one to have it)

src/environments/environment.ts
```diff
@@ -31,11 +31,11 @@ export const environment = {
     }
   },
   VISA_CHECKOUT_URLS: {
-    DEV_BUTTON_URL: 'https://sandbox.secure.checkout.visa.com/wallet-services-web/xo/button.png',
-    DEV_SDK: 'https://sandbox-assets.secure.checkout.visa.com/checkout-widget/resources/js/integration/v1/sdk.js',
+    LIVE_BUTTON_URL: 'https://secure.checkout.visa.com/wallet-services-web/xo/button.png',
```
aaron-moore-ST: to explain this change, 'live' and 'test' are the terms we use on our production system so we wanted them the same to make things easier

src/core/integrations/CardinalCommerce.ts
```diff
@@ -166,9 +166,8 @@ export class CardinalCommerce {
     if (this._isCardEnrolledAndNotFrictionless(responseObject)) {
       this._authenticateCard(responseObject);
     } else {
-      this._authorizePayment({
-        parenttransactionreference: responseObject.transactionreference
-      });
+      this._threedQueryTransactionReference = responseObject.transactionreference;
```
aaron-moore-ST: this change is because parenttransactionreference is overridden later on with _threedQueryTransactionReference, so we can reuse that code instead.

src/core/integrations/CardinalCommerce.ts
```diff
@@ -166,9 +166,8 @@ export class CardinalCommerce {
...
+      this._authorizePayment({});
```
aaron-moore-ST: perhaps this should default the transactiondata to an empty object, it's a minor change but would make this prettier

james-simpson-ST: Done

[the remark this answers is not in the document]
james-simpson-ST: Done

src/core/shared/Frame.ts
james-simpson-ST: Done

PR #168, opened 2019-05-14
ST-181 Card numnber formatting new pull request

pgs-gmartowski:
New PR for ST-181 cause of deleting `js-payments-testing` and can't restore that one.


pgs-gmartowski:
> Great work lgtm
> 
> However me and Aaron have been discussing a naming convention for unit tests and think we should use
> 
> ClassName.func_name
> 
> E.g. instead of "Method getPossibleCardLength" -> "CardNumber.getPossibleCardLength"
> 
> I've updated the tests for the classes I tested on my unit test coverage branch could you apply it to this class and then as we make changes to others we'll update the rest over time?
> 
> But other than that this looks good to go :)

Sure, good point !
I changed all the test suites descriptions in CardNumber and if there'll be an opportunity I change the rest of the in other classes.

PR #169, opened 2019-05-14
Feature/st 197 allow locale to be passed in

james-simpson-ST:
Example.ts can now read JWT from URL GET parameters so Behavioural tests can pass in a JWT with locale set to fr_FR to test translations e.g.:

https://localhost:8443?jwt=<new test jwt>


example/index.ts
```diff
@@ -9,13 +9,17 @@ import './style.scss';
...
+  const applePayVisaCheckoutJwt =
+    'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJsaXZlMl9hdXRvand0IiwiaWF0IjoiMTU1NjE5MjQxMDc0OSIsInBheWxvYWQiOnsiYmFzZWFtb3VudCI6IjEwMDAiLCJjdXJyZW5jeWlzbzNhIjoiR0JQIiwic2l0ZXJlZmVyZW5jZSI6ImxpdmUyIiwiYWNjb3VudHR5cGVkZXNjcmlwdGlvbiI6IkVDT00ifX0.YPXFxhb72eEc5yUihniXmbyVEQfPtvULk9W36uKR2zg';
+
+  const jwt = parsedUrl.searchParams.get('jwt') || cardinalJwt;
```
pgs-gmartowski: if it's in the example file (example/index.ts) IMO it shouldn't have any js methods , only raw data; shouldn't we add this as some static function in ST object ?

aaron-moore-ST: I don't think this is even part of the ST object, it's purely a part of the example page.

james-simpson-ST:
I will leave this in example.ts for now as its just needed for the example page

PR #171, opened 2019-05-15
ST-201 Bugfix - mocks

pgs-gmartowski:


src/core/integrations/ApplePay.ts
```diff
@@ -338,7 +338,10 @@ class ApplePay {
     if (walletsession) {
       this.merchantSession = JSON.parse(walletsession);
       this.validateMerchantRequestData.walletmerchantid = this.merchantSession.merchantIdentifier;
-      this.session.completeMerchantValidation(this.merchantSession);
+      this.setNotification(NotificationType.Success, Language.translations.PAYMENT_SUCCESS);
+      if (this.session) {
+        this.session.completeMerchantValidation(this.merchantSession);
```
aaron-moore-ST: It looks like this would display a success message but then silently fail to continue if the session doesn't exist, does onValidateMerchantResponseFailure need to be called if the session didn't exist here?

src/core/integrations/ApplePay.ts
```diff
@@ -338,7 +338,10 @@ class ApplePay {
     if (walletsession) {
       this.merchantSession = JSON.parse(walletsession);
       this.validateMerchantRequestData.walletmerchantid = this.merchantSession.merchantIdentifier;
-      this.session.completeMerchantValidation(this.merchantSession);
+      this.setNotification(NotificationType.Success, Language.translations.PAYMENT_SUCCESS);
```
aaron-moore-ST: onValidateMerchant doesn't authorise the payment so this message is misleading. Do we want to generate a message here? If we do we should work out a more informative message.

src/core/integrations/ApplePay.ts
pgs-gmartowski: this was dedicated for testing purposes (if we launch tests on some other browsers than Safari) but actually this test case doesn;t make sense.
I deleted this line.

src/core/integrations/ApplePay.ts
pgs-gmartowski: this can be sth that has left after merging - Marek reported this as sth to be deleted.
Deleted.

src/core/integrations/ApplePay.ts
```diff
@@ -336,12 +336,9 @@ class ApplePay {
   public onValidateMerchantResponseSuccess(response: any) {
     const { walletsession } = response;
     if (walletsession) {
-      this.merchantSession = JSON.parse(walletsession);
+      this.merchantSession = JSON.parse(response.walletsession);
```
aaron-moore-ST: since this line and the above are both using walletsession should we either have both use const walletsession or both use response.walletsession and remove the temporary variable?

pgs-gmartowski: true, I deleted this const and leave just the response object.

PR #172, opened 2019-05-16
ST-26 handling errors from backend

pgs-gmartowski:


pgs-gmartowski:
I leave here  James' comment from ST-141 PR:

> I have been looking into this and from a visual point of view this is fine when the button locks but I have the following issues:

If the payment fails the form needs to unlock again e.g. if the AUTH request declines
Once the AUTH request is successful it'd be nice if we got rid of the Please wait and perhaps loaded in a success message like Payment complete
I have noticed that if the button is clicked then the form/button is locked we also need to do this if any of the inputs have the enter key hit with this I think the below change should take place:
Instead of this being triggered inside the Form class it needs to move into the Payment class. Then it needs to use the messageBus to send a public message to the parent that will trigger the preloader. The reason for the messageBus is that now we have some requests (for Cardinal) that happen inside the control frame and some that happen (e.g. for ApplePay/VisaCheckout) that happen in the parent frame. If we use the message bus we can guarentee it gets fired by both. (We should also check that while we're processing a VisaCheckout the form is locked too)

Let me know if more explanation is needed

james-simpson-ST:
Just tried this out and have a few issues I think

I just tried an AUTH which should decline (4000 0000 0000 0002) and it seems to return errorcode 70000 errormessage "Decline" but the success message was shown

Also the button does lock the form but if I hit enter into any of the form fields we still perform the submit - when the button is locked it should also prevent any kind of submit not just on the button

james-simpson-ST:
Code lgtm as long as pipeline is passing :)

PR #182, opened 2019-05-20
Feature/stjs 100 cc automated tests

james-simpson-ST:
This PR improves the Cardinal commerce mock so that tests can pass on Behavioural tests

Note I have noticed some refactoring that is needed between Visa/ApplePay and Cardinal (in the control frame) but I feel it's a separate task to do this so I have created STJS-205 to do this refactoring


james-simpson-ST:
Note before we pull this back in we should also change the branch in Dockerfile back to develop but this is waiting on PR https://github.com/SecureTrading/js-payments/pull/182

PR #183, opened 2019-05-21
Feature/st 73 define paying options

james-simpson-ST:


example/index.ts
```diff
@@ -3,7 +3,7 @@
  * This is code fired on merchant's side.
  * It can be treated as a reference for merchants how to integrate with STJS.
  */
-import { ST } from '../src/stjs';
+import SecureTrading from '../src/ST';
```
aaron-moore-ST: While we're here is it worth fixing our _other_ problem? Otherwise we're going to need to update the hosted example page again when this hits the branch.

example/index.ts
```diff
@@ -20,55 +20,40 @@ import './style.scss';
...
-          merchantId: 'SDUT1MEXJO10RARJF2S521ImTyKfn3_JmxePdXcydQIUb4kx4',
-          paymentRequest: { subtotal: '20.00' },
-          placement: 'st-visa-checkout',
-          settings: { displayName: 'My Test Site' }
-        }
-      }
-    ]
-  );
+    placement: 'st-apple-pay',
+    sitesecurity: 'gABC123DEFABC' // TODO this shouldn't be needed should it?
```
aaron-moore-ST: Since I believe we should now be using the gateway rather than wallet listener I agree that we should no longer need the sitesecurity field.

src/ST.ts
```diff
@@ -0,0 +1,70 @@
+import Cards from './core/classes/CardFrames.class';
```
aaron-moore-ST: Weird that this is an add and delete rather than a move. I wonder if we could rebase to fix that as it would keep the history of the file intact?

src/core/classes/CardFrames.class.ts
```diff
@@ -1,15 +1,14 @@
...
-import { StJwt } from '../shared/StJwt';
 import { IStyles } from '../shared/Styler';
+import Register from './RegisterFrames.class';
 
 /**
- * Defines all elements of form and their  placement on merchant site.
+ * Defines all card elements of form and their  placement on merchant site.
  */
-class Form {
+class Form extends Register {
```
aaron-moore-ST: Should the Form class be renamed along with the containing module?

src/core/classes/CommonFrames.class.ts
```diff
@@ -0,0 +1,84 @@
...
+import DomMethods from '../shared/DomMethods';
+import MessageBus from '../shared/MessageBus';
+import Selectors from '../shared/Selectors';
+import { IStyles } from '../shared/Styler';
+import Register from './RegisterFrames.class';
+
+/**
+ * Defines all non field elements of form and their placement on merchant site.
+ */
+class Form extends Register {
```
aaron-moore-ST: Is it misleading that CardFrames and Common frames both contain Form (but it means different things)?

src/core/classes/Wallet.class.ts
```diff
@@ -1,59 +0,0 @@
-import { environment } from '../../environments/environment';
```
aaron-moore-ST: I'm not sure I see where some of the content of this file ended up, what replaces it?

james-simpson-ST: It was an unused import

src/ST.ts
james-simpson-ST: I am not sure and don't have the necessary git skills at the moment to fix it and with the fact it has completely changed I don't think its worth it for now

src/core/classes/CardFrames.class.ts
james-simpson-ST: Yup I'll update

src/core/classes/CommonFrames.class.ts
james-simpson-ST: Yup I'll update

src/core/classes/Wallet.class.ts
james-simpson-ST: It has moved into ST.js or is no longer needed (i.e. the splitting of the config isn't needed because the config itself is passed into either ApplePay or VisaCheckout now rather than a list of configs like before

example/index.ts
james-simpson-ST: Good I'll try and remove

example/index.ts
james-simpson-ST: I will see how much effort is involved and if its small I will do it now, if not I might have to leave it as I don't want to increase the scope too much on this task with the time constraints

james-simpson-ST: Done

example/index.ts
james-simpson-ST: Done

src/core/classes/CardFrames.class.ts
james-simpson-ST: Done

src/core/classes/CommonFrames.class.ts
james-simpson-ST: Done

PR #185, opened 2019-05-27
ST-167 Polyfills for URL and URLSearchParams

pgs-gmartowski:


james-simpson-ST:
Does the polyfill automatically get applied as soon as it's included in the dependencies? I wasn't sure if it would unless we were using webpack.ProvidePlugin

example/index.html
```diff
@@ -93,7 +93,7 @@ <h1 class="example-form__title">
         var jwt = parsedUrl.searchParams.get('jwt') || cardinalJwt;
 
         var st = SecureTrading({
-          jwt,
+          jwt: jwt, // because of IE :/
```
james-simpson-ST: Probably wants to be more descriptive like the following if it's in our example:

// Can't use property shorthand because it isn't supported by IE

pgs-gmartowski: yes, that was not accurate comment :)

PR #195, opened 2019-05-31
Bugfix/ie compatibility

james-simpson-ST:


.babelrc
```diff
@@ -1,15 +0,0 @@
-{
```
pgs-gmartowski: why we should now have babel.config.js instead of .babelrc ?

james-simpson-ST: By using babel.config.js the babel transpiling now works on examples as well as the src directory https://babeljs.io/docs/en/config-files#project-wide-configuration

PR #196, opened 2019-06-03
Bugfix/st 186 fix tab index on firefox

james-simpson-ST:


james-simpson-ST:
> Code looks good, however I'm not sure if UX is correct now - it looks like cursor is 'trapped' again in single iframe.
> I checked if my Firefox is out of date, but it's not - 61.0.2

I will try again on my machine, just to make sure I understand, does the cursor get 'trapped' in the first field once you tab into it where you can't tab back out of it?

james-simpson-ST:
@pgs-gmartowski I have been unable to reproduce the issue of getting 'trapped' in a field - perhaps a demo might be needed

pgs-gmartowski:
> > Code looks good, however I'm not sure if UX is correct now - it looks like cursor is 'trapped' again in single iframe.
> > I checked if my Firefox is out of date, but it's not - 61.0.2
> 
> I will try again on my machine, just to make sure I understand, does the cursor get 'trapped' in the first field once you tab into it where you can't tab back out of it?

Exactly; it only happens on Firefox, I'd tried to update my browser and find the reason on my PC but it looks like browsers issue. 

I'll ask Marek if he also has such problem.

pgs-gmartowski:
> > > Code looks good, however I'm not sure if UX is correct now - it looks like cursor is 'trapped' again in single iframe.
> > > I checked if my Firefox is out of date, but it's not - 61.0.2
> > 
> > 
> > I will try again on my machine, just to make sure I understand, does the cursor get 'trapped' in the first field once you tab into it where you can't tab back out of it?
> 
> Exactly; it only happens on Firefox, I'd tried to update my browser and find the reason on my PC but it looks like browsers issue.
> 
> I'll ask Marek if he also has such problem.

@james-simpson-ST 
The reason was that I had 32-bit version of Firefox and after updating to the 64-bit everything is fine.

PR #207, opened 2019-06-04
Feature/st 74 define payment types

james-simpson-ST:


src/core/shared/Selectors.ts
```diff
@@ -37,7 +37,7 @@ export default class Selectors {
   public static CONTROL_FRAME_COMPONENT: string = `${environment.FRAME_URL}/control-frame.html`;
   public static readonly ANIMATED_CARD_COMPONENT: string = `${environment.FRAME_URL}/animated-card.html`;
 
-  public static readonly MERCHANT_FORM_SELECTOR: string = 'st-form';
+  public static MERCHANT_FORM_SELECTOR: string = 'st-form';
```
pgs-gmartowski: if this class field might be changed it'll be better to change it into non-static

test/components/animated-card/AnimatedCard.spec.ts
```diff
@@ -222,6 +303,47 @@ describe('Class AnimatedCard', () => {
...
+    it('should get all configs', () => {
+      // @ts-ignore
+      instance._params = { defaultPaymentType: 'AMEX', paymentTypes: 'VISA,MASTERCARD,AMEX' };
+      // @ts-ignore
+      expect(instance.getBinLookupConfig()).toStrictEqual({
+        defaultCardType: 'AMEX',
+        supported: ['VISA', 'MASTERCARD', 'AMEX']
+      });
+    });
+  });
```
pgs-gmartowski: I also wrote some UT in ST-180 for Animated Card :) 
Maybe we could merge them after ST-74 will be in develop.

src/core/shared/Selectors.ts
james-simpson-ST: Done

PR #210, opened 2019-06-05
ST-221 Highlight merchant fields

pgs-gmartowski:


src/components/card-number/CardNumber.ts
```diff
@@ -168,16 +173,6 @@ export default class CardNumber extends FormField {
...
-  /**
-   * Checks the response from backend
-   */
-  public backendValidation() {
-    this._messageBus.subscribe(MessageBus.EVENTS.VALIDATE_CARD_NUMBER_FIELD, (data: any) => {
-      this.checkBackendValidity(data);
-      this.validation.validate(this._inputElement, this._messageElement);
-    });
-  }
-
```
pgs-gmartowski: info: I moved all validation methods to Validation.ts class

james-simpson-ST: Nice refactor :)

src/core/classes/MerchantFields.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /**
+   * Checks if Merchant's form has inputs with data-st-name and  returns them.
+   */
+  public findAllMerchantInputs() {
+    const merchantFieldsNamesArray = [];
+    for (let i = 0; i < this._merchantInputs.length; ++i) {
+      // @ts-ignore
+      if (this._merchantInputs[i].dataset.hasOwnProperty(MerchantFields.DATA_ATTRIBUTE_NAME)) {
```
james-simpson-ST: I tried to use dataset for parsing the merchant fields and its not supported in IE so I had to use hasAttribute/getAttribute/setAttribute so please update to use them

james-simpson-ST: Might only be in IE10 which we haven't added to the pipeline yet :(

pgs-gmartowski: yes, it looks like it's only partial support on IE10 :/ Ok, I'll change that.

src/components/card-number/CardNumber.ts
pgs-gmartowski: :)

src/core/classes/MerchantFields.ts
pgs-gmartowski: done

PR #211, opened 2019-06-06
Feature/st 216 allow endpoint to be overridden

james-simpson-ST:


test/core/classes/StTransport.spec.ts
```diff
@@ -43,7 +44,7 @@ describe('StTransport class', () => {
       );
       await st.sendRequest(requestObject);
       expect(st.fetchRetry).toHaveBeenCalledTimes(1);
-      expect(st.fetchRetry).toHaveBeenCalledWith(StTransport.GATEWAY_URL, {
+      expect(st.fetchRetry).toHaveBeenCalledWith('https://example.com', {
```
pgs-gmartowski: why not `DEFAULT_PARAMS.gatewayUrl` ?

james-simpson-ST: Done :)

PR #212, opened 2019-06-09
Feature/st 207 applepay code review

james-simpson-ST:
From review:
- ifApplePayIsAvailable changed to return true/false
- ApplePay should return an error response not a success response if the AUTH does not successfully process
- VisaCheckoutMock has been refactored to mock less and use more of the VisaCheckout class

### DIFF
--- a/src/core/integrations/VisaCheckout.ts
+++ b/src/core/integrations/VisaCheckout.ts
@@ -1,4 +1,3 @@
-declare const V: any;
 import { environment } from '../../environments/environment';
 import { INotificationEvent, NotificationType } from '../models/NotificationEvent';
 import MessageBus from '../shared/MessageBus';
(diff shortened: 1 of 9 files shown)


src/core/integrations/VisaCheckout.ts
```diff
@@ -217,16 +215,54 @@ export class VisaCheckout {
...
+    });
+    V.on(VisaCheckout.VISA_PAYMENT_RESPONSE_TYPES.ERROR, () => {
+      this._onError();
+    });
+    V.on(VisaCheckout.VISA_PAYMENT_RESPONSE_TYPES.CANCEL, () => {
+      this._onCancel();
+    });
+  }
+
+  private setConfiguration = (config: any, settings: any) => (settings || config ? { ...config, ...settings } : {});
```
pgs-gmartowski: `config` and `settings` parameters should be described by interfaces

james-simpson-ST: I have added a basic interface for both of these however as they will be used by the Visa config which will be out of our control I haven't described a complex interface only providing a basic key/value interface

PR #246, opened 2019-06-18
feature/ST-226-security-code-UT

pgs-gmartowski:


test/components/security-code/SecurityCode.spec.ts
```diff
@@ -166,16 +182,37 @@ describe('SecurityCode', () => {
...
 
   // given
-  describe('onPaste', () => {
-    const { instance } = securityCodeFixture();
+  describe('onPaste()', () => {
+    const { instance } = SecurityCodeFixture();
+    let spyIsMaxLengthReached: jest.SpyInstance;
+    let spySendState: jest.SpyInstance;
+    // @ts-ignore
+    // @TODO clipboard event is not recognizable.
```
pgs-gmartowski: @james-simpson-ST Possibly we need to mock Clipboard event cause typescript doesn't properly implemented it (that's only my thought be it looks like that). What I receive in error logs it's ClipboardEvent is not defined.

pgs-gmartowski: I just left this case for a moment because of new tasks from todays DS.

pgs-gmartowski: same situation is in FormFields.ts and other classes connected with input components

james-simpson-ST: Yes I think the other tasks are higher priority so that makes sense :)

PR #248, opened 2019-06-18
bugfix/ST-247-missingtranslation-for-pay-button

pgs-gmartowski:


src/core/classes/CardFrames.class.ts
```diff
@@ -30,12 +31,14 @@ export default class CardFrames extends RegisterFrames {
    * @param disabledState
    * @private
    */
-  private static _setSubmitButtonProperties(element: any, disabledState: boolean) {
+  private _setSubmitButtonProperties(element: any, disabledState: boolean) {
+    console.log(this._translator.translate(Language.translations.PAY));
```
james-simpson-ST: Need to remove log lines

PR #249, opened 2019-06-19
feature/ST-258-include-cardinal-data-in-error-responses

pgs-gmartowski:


src/core/integrations/CardinalCommerce.ts
```diff
@@ -29,39 +29,25 @@ export interface IThreeDQueryResponse {
  * 6.Cardinal.on('pauments.validated) - process auth or return failure
  */
 export class CardinalCommerce {
-  get tokenise(): boolean {
-    return this._tokenise;
-  }
-
-  set tokenise(value: boolean) {
```
pgs-gmartowski: info: there is no need to use getter / setter

src/core/integrations/CardinalCommerce.ts
```diff
@@ -29,39 +29,25 @@ export interface IThreeDQueryResponse {
...
     INIT: 'init',
     SETUP_COMPLETE: 'payments.setupComplete',
     VALIDATED: 'payments.validated'
   };
-  private static VALIDATION_EVENTS = {
-    ERROR: 'ERROR',
-    FAILURE: 'FAILURE',
-    NOACTION: 'NOACTION',
-    SUCCESS: 'SUCCESS'
-  };
```
pgs-gmartowski: info: redundant, we've got Selectors class for that

src/core/integrations/CardinalCommerce.ts
```diff
@@ -1,4 +1,13 @@
 import { environment } from '../../environments/environment';
+import {
+  IAuthorizePaymentResponse,
+  IOnCardinalValidated,
+  IThreeDInitResponse,
+  IThreeDQueryResponse,
+  ON_CARDINAL_VALIDATED_STATUS,
+  PAYMENT_BRAND,
+  PAYMENT_EVENTS
+} from '../models/CardinalCommerce';
```
pgs-gmartowski: info: I did small refactor, some of the static fields are moved to the CardinalCommerce model

src/core/integrations/CardinalCommerce.ts
```diff
@@ -165,59 +87,175 @@ export class CardinalCommerce {
...
-   * @protected
    */
   protected _onCardinalLoad() {
     Cardinal.configure(environment.CARDINAL_COMMERCE.CONFIG);
-    Cardinal.on(CardinalCommerce.PAYMENT_EVENTS.SETUP_COMPLETE, () => {
+    Cardinal.on(PAYMENT_EVENTS.SETUP_COMPLETE, () => {
       this._onCardinalSetupComplete();
     });
-    Cardinal.on(CardinalCommerce.PAYMENT_EVENTS.VALIDATED, (data: any, jwt: any) => {
+    Cardinal.on(PAYMENT_EVENTS.VALIDATED, (data: IOnCardinalValidated, jwt: string) => {
```
pgs-gmartowski: info: I've got rid of the `any` types from this file - eventually we want to have all inputs/outputs defined.

src/core/models/CardinalCommerce.ts
```diff
@@ -0,0 +1,67 @@
...
+interface IThreeDInitResponse {
+  cachetoken: string;
+  errorcode: string;
+  errormessage: string;
+  requesttypedescription: string;
+  threedinit: string;
+  transactionstartedtimestamp: string;
+}
+
+interface IThreeDQueryResponse {
```
pgs-gmartowski: info: this type of response was to long to place it in CC class, that's why I moved it here

james-simpson-ST: The stgateway could be updated in the future to return another field without warning, can we not just use the following for any field we don't care if it is or isn't returned?

[field: string]: string;

src/core/integrations/CardinalCommerce.ts
james-simpson-ST: Agreed :)

james-simpson-ST:
Code lgtm with some nice clean ups :)

But unit tests are failing the pipeline

src/core/models/CardinalCommerce.ts
pgs-gmartowski: ok, so I should leave only those fields of this interface which were in the previous version ? It looked like that:
```
export interface IThreeDQueryResponse {
  acquirertransactionreference: string;
  acsurl: string;
  enrolled: string;
  threedpayload: string;
  transactionreference: string;
}
```

pgs-gmartowski: restored previous interface

PR #278, opened 2019-07-03
Bugfix - wrong translation after luhn check

pgs-gmartowski:


pgs-gmartowski:
@james-simpson-ST On Monday I changed validation method and some translations has changed - here is fix for this issue.

PR #284, opened 2019-07-04
ST-205 refactor Payment.ts class

pgs-gmartowski:


src/core/shared/Payment.ts
```diff
@@ -59,7 +85,15 @@ export default class Payment {
       merchantData,
```
pgs-gmartowski: @james-simpson-ST is `termurl: 'https://termurl.com'` still needed ?

src/core/integrations/ApplePay.ts
```diff
@@ -134,7 +135,7 @@ export class ApplePay {
    * If yes, returns ApplePaySession object, if not returns undefined.
    */
   public ifApplePayIsAvailable() {
-    return ApplePaySession ? true : false;
+    return !!ApplePaySession;
```
pgs-gmartowski: it was tslint hint - it really looks less obscure than ternary above

src/core/shared/Payment.ts
james-simpson-ST: Last time I checked this had not been done by the gateway team :( I will re-raise it with them

src/core/integrations/ApplePay.ts
james-simpson-ST: Makes sense to me

PR #288, opened 2019-07-09
bugfix/ST 290 invalid field validation message instead of pattern mismatch

pgs-gmartowski:


src/core/shared/Validation.ts
```diff
@@ -164,7 +164,7 @@ export default class Validation extends Frame {
    */
   public setError(inputElement: HTMLInputElement, messageElement: HTMLElement, message: string) {
     inputElement.classList.add(Validation.ERROR_FIELD_CLASS);
-    if (messageElement) {
+    if (messageElement && messageElement.innerText !== 'Value mismatch pattern') {
```
pgs-gmartowski: info: yeah, I know it's really ugly, I'll change it as soon as I am back home.

src/core/shared/Validation.ts
```diff
@@ -169,7 +170,7 @@ export default class Validation extends Frame {
...
```
james-simpson-ST: Can't this use Language.transactions.VALIDATION_ERROR_PAYYERN_MISMATCH instead of 'Value mismatch pattern'

pgs-gmartowski: Good point ! Fixed

PR #305, opened 2019-07-11
ST-240 Import completed translations

james-simpson-ST:
Imported translations from 3rd party translation service


src/core/shared/Language.ts
```diff
@@ -18,7 +18,7 @@ class Language {
     PAYMENT_CANCELLED: 'Payment has been cancelled',
     PAYMENT_ERROR: 'An error occurred',
     PAYMENT_SUCCESS: 'Payment has been successfully processed',
-    PROCESSING: 'Processing',
+    PROCESSING: 'Processing...',
```
pgs-gmartowski: I remember that dots make translation does not work, I'll check it out

src/core/translations/da_DK.json
```diff
@@ -1,24 +1,24 @@
...
-  "Security code": "Security code",
-  "Ok": "Payment has been successfully processed",
-  "Method not implemented": "Method not implemented",
-  "Form is not valid": "Form is not valid",
-  "Pay": "Pay",
-  "Processing": "Processing...",
-  "Invalid field": "Invalid field",
-  "Card number is invalid": "Card number is invalid"
+  "Field is required": "Feltet skal udfyldes",
+  "An error occurred": "Der er opst�et en fejl",
```
pgs-gmartowski: some of the letters are replaced by �, however in German **ü** is displaying correctly. I googled that in Danish it might be **Ø**

PR #309, opened 2019-07-16
feature/ST-160-control-frame-ut

pgs-gmartowski:


webpack.common.js
```diff
@@ -142,7 +142,7 @@ module.exports = {
           {
             loader: 'tslint-loader',
             options: {
-              emitErrors: true
+              emitErrors: false
```
james-simpson-ST: This wants to be set back to true

pgs-gmartowski: yes, I always forget to change that back ;)

pgs-gmartowski: done

PR #324, opened 2019-07-18
ST-245 Merchant Fields UT improvement

pgs-gmartowski:


src/components/card-number/CardNumber.ts
```diff
@@ -35,6 +35,7 @@ export default class CardNumber extends FormField {
     this.isCardNumberValid = true;
     this.cardNumberLength = CardNumber.STANDARD_CARD_LENGTH;
     this.setFocusListener();
+    this.setBlurListener();
```
pgs-gmartowski: info: setting blur listener fixed the problem - on ios blur event is not triggered, so the validation was not working

PR #341, opened 2019-07-25
feature/ST-26-implement-event-logging-via-google-analytics

pgs-gmartowski:


src/core/integrations/GoogleAnalytics.ts
```diff
@@ -0,0 +1,115 @@
+import DomMethods from '../shared/DomMethods';
+
+/**
+ * Creates HTML markups <script> and add Google Analytics source to it.
+ */
+class GoogleAnalytics {
+  private static GA_MEASUREMENT_ID: string = '42057093-5';
```
james-simpson-ST: Can this be set to have a different value for test/live using livestatus like we have for VisaCheckout in _setLiveStatus to use a different script?

For test use: "42057093-5" and for live use: "42057093-6"

src/core/integrations/GoogleAnalytics.ts
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * Adds all required features by interpolating static strings.
+   * @private
+   */
+  private static _returnScriptWithFeatures() {
+    return `${GoogleAnalytics.GA_INIT_SCRIPT_CONTENT}
+    ${GoogleAnalytics.GA_DISABLE_COOKIES}
+    ${GoogleAnalytics.GA_IP_ANONYMIZATION}
+    ${GoogleAnalytics.GA_DISABLE_ADVERTISING_FEATURES}
+    ${GoogleAnalytics.GA_PAGE_VIEW}`;
```
james-simpson-ST: Currently we only get one event (when the page loads) we also need to log additional events at different stages in the transaction e.g.

Cardinal init completed
Cardinal threedquery completed
Cardinal payment.validated
Cardinal auth completed

Similarly for ApplePay/VisaCheckout 

I believe the api for this is 
ga('send', {
  hitType: 'event',
  eventCategory: 'Cardinal',
  eventAction: 'init',
  eventLabel: 'Cardinal Init Completed'
});

src/core/integrations/GoogleAnalytics.ts
```diff
@@ -0,0 +1,115 @@
+import DomMethods from '../shared/DomMethods';
```
james-simpson-ST: I have updated implementation as the events were not appearing in GA dashboard. I found this useful when debugging:

https://developers.google.com/analytics/devguides/collection/analyticsjs/debugging

pgs-gmartowski: wow, cool stuff !
 I've never used that before

src/core/integrations/GoogleAnalytics.ts
pgs-gmartowski: yup, good point; I added `GA_SCRIPT_SRC` and `GA_MEASUREMENT_ID` to environment files

src/core/integrations/GoogleAnalytics.ts
pgs-gmartowski: ok, I think I need couple of minutes more for that - I'm moving ST-36 to 'in progres' again.

pgs-gmartowski: @james-simpson-ST I've added `send` method in couple of places in each class (CC, VC and AP), but I'm not sure if there is enough or there need to be some more events to be tracked ?

src/core/integrations/GoogleAnalytics.ts
```diff
@@ -0,0 +1,122 @@
...
+  private static GA_MEASUREMENT_ID: string = environment.GA_MEASUREMENT_ID;
+  private static GA_INIT_SCRIPT_CONTENT: string = `window.ga=window.ga||function(){(ga.q=ga.q||[]).
+  push(arguments)};ga.l=+new Date;
+`;
+  private static GA_SCRIPT_SRC: string = environment.GA_SCRIPT_SRC;
+  private static GA_DISABLE_COOKIES: string = `ga('create', 'UA-${GoogleAnalytics.GA_MEASUREMENT_ID}'
+  , {'storage': 'none'});`;
+  private static GA_IP_ANONYMIZATION: string = `ga('set', 'anonymizeIp', true);`;
+  private static GA_DISABLE_ADVERTISING_FEATURES: string = `ga('set', 'allowAdFeatures', false);`;
+  private static GA_PAGE_VIEW: string = `ga('send', 'pageview');`;
```
james-simpson-ST: Currently by default the following line will send the page we're using to GA, but it will include the query parameters
`private static GA_PAGE_VIEW: string = `ga('send', 'pageview');`;` 

As this will be the merchants page we don't know what these query parameters could contain (it could be some persional identifiable information) so we should change this to send only the path without the query parameters

I think this should be achieveable if we modify the path sent https://developers.google.com/analytics/devguides/collection/analyticsjs/pages

src/core/integrations/GoogleAnalytics.ts
james-simpson-ST: I think this is enough and will mean it is easy to add more in future if we need to :)

src/core/integrations/GoogleAnalytics.ts
pgs-gmartowski: true, so I think this should do the work:
`private static GA_PAGE_VIEW: string = `ga('send', 'pageview', location.pathname);`;`
According to documentation this will only send the URL without query string (previously we were sending 'title', `location` and `page`).

PR #345, opened 2019-07-31
bugfix/ST-300-processing-button-is-not-translated

pgs-gmartowski:


src/core/classes/CardFrames.class.ts
```diff
@@ -49,6 +51,8 @@ class CardFrames extends RegisterFrames {
     this.onInit();
     this._translator = new Translator(this.params.locale);
     this._getSubmitButton();
+    this._payMessage = this._translator.translate(Language.translations.PAY);
+    this._processingMessage = `${this._translator.translate(Language.translations.PROCESSING)} ...`;
```
pgs-gmartowski: info: these two must be translated in constructor because in `_setSubmitButtonProperties` it's unfortunately too late :/

PR #354, opened 2019-07-31
bugfix/ST-302-all-fields-are-highlighted-if-one-of-them-has-value-mismatch-pattern-error

pgs-gmartowski:


src/core/shared/Validation.ts
```diff
@@ -312,7 +312,6 @@ export default class Validation extends Frame {
    * @private
    */
   private _assignErrorDetails(inputElement: HTMLInputElement, messageElement: HTMLElement, message: string) {
-    inputElement.classList.add(Validation.ERROR_FIELD_CLASS);
```
pgs-gmartowski: info: looks like as simple as that ;)

PR #364, opened 2019-08-07
feature/ST-296-handle-response-JWT

pgs-gmartowski:


src/core/shared/Payment.ts
```diff
@@ -2,6 +2,7 @@ import { IStRequest } from '../classes/StCodec.class';
 import StTransport from '../classes/StTransport.class';
```
james-simpson-ST: I think some confusion has happened here (too many JWTs)

The cardinalCommerceCacheToken does not need to change

Then in StCodec.class.ts we need to change in decode() to replace this._jwt with decoded.jwt so we end up with this in the middle of the decode() function
```javascript
const decoded = StCodec._decodeResponseJwt(responseData.jwt, reject);
this._jwt = decoded.jwt
```

Additionally in CardinalCommerce._authorizePayment the following line shouldn't be needed anymore
`data.parenttransactionreference = this._threedQueryTransactionReference; `

pgs-gmartowski: ok, I see. but this implementation in Payment.ts is still needed, right ?: 
`
public threeDInitRequest() {
    return this._stTransport.sendRequest(this._threeDInitRequestBody).then((result: { jwt: string; response: any }) => {
      const {
        payload: { jwt, response }
      } = new StJwt(result.jwt);
      const threeDInitResult = { jwt, response: response[0] };
      // @ts-ignore
      this._cardinalCommerceCacheToken = result.response.cachetoken;
      return threeDInitResult;
    });
  }
`

james-simpson-ST: Yes I believe so

src/core/classes/StCodec.class.ts
```diff
@@ -277,6 +277,7 @@ class StCodec {
       if ('json' in responseObject) {
         responseObject.json().then(responseData => {
           const decoded = StCodec._decodeResponseJwt(responseData.jwt, reject);
+          this._jwt = decoded.jwt;
```
james-simpson-ST: There appears to be an issue with scope here, because we are in a promise we do not have access to "this"

Also I believe I said the wrong thing earlier where it should be `decoded.payload.jwt` (sorry about that)

james-simpson-ST: We get:
Uncaught (in promise) TypeError: Cannot set property '_jwt' of undefined

Also can you manually verify this is working once updated to ensure you can still process a successful card payment

pgs-gmartowski: damn, stupid me, yes your're right, `this` refers to different scope here :/

PR #365, opened 2019-08-07
feature/ST-306-add-merchant-callback

pgs-gmartowski:


example/index.html
```diff
@@ -102,7 +102,21 @@ <h1 class="example-form__title">
...
+          submitOnSuccess: false,
+          submitCallback: function someFancyfunction(data) {
+            var promise = new Promise(function(resolve, reject) {
+              setTimeout(function() {
+                resolve('foo');
+              }, 300);
+            });
+            promise
+              .then(() => {
+                console.error('success');
```
pgs-gmartowski: info: this is only the example to check if this callback is working fine

src/core/classes/CommonFrames.class.ts
```diff
@@ -10,9 +10,6 @@ import RegisterFrames from './RegisterFrames.class';
  * Defines all non field elements of form and their placement on merchant site.
  */
 class CommonFrames extends RegisterFrames {
-  get merchantForm(): any {
```
pgs-gmartowski: info: this is not used anymore

src/core/classes/CommonFrames.class.ts
```diff
@@ -130,11 +130,11 @@ class CommonFrames extends RegisterFrames {
   }
 
   /**
-   * _isThreedComplete
+   * Checks enrolled ( _isCardEnrolledAndNotFrictionless ) card with request type and returns boolean.
    * @param data
    * @private
    */
-  private _isThreedComplete(data: any) {
+  private _isThreedComplete(data: any): boolean {
```
pgs-gmartowski: info: just added some Typescript details and typedocs to make class more neat

src/core/classes/CommonFrames.class.ts
```diff
@@ -172,11 +173,14 @@ class CommonFrames extends RegisterFrames {
...
   /**
-   * _onTransactionComplete
+   * Function which is triggered each time transaction has been completed.
+   * It calls submitCallback specified (or not by merchant) and submits or not
+   * (also specified by merchant) form as well.
    * @param data
    * @private
    */
   private _onTransactionComplete(data: any) {
+    this._submitCallback(data);
```
pgs-gmartowski: should work now - I checked it id couple of configuration (with submitOnSuccess, without it etc ) and it works fine

src/core/classes/CommonFrames.class.ts
james-simpson-ST: Nice improvements :)

test/core/classes/CommonFrames.spec.ts
```diff
@@ -215,6 +215,8 @@ describe('CommonFrames', () => {
       // @ts-ignore
       instance._shouldSubmitForm = jest.fn().mockReturnValueOnce(true);
       // @ts-ignore
+      instance._submitCallback = jest.fn();
```
james-simpson-ST: Should we unit test that this has been called?

james-simpson-ST:
Just trying this manually and need a couple of changes actually:

IE11 doesn't work with the callback you've put in (can it just be a simpler function?

Also this gets fired on every response from the gateway instead of just the final result so we actually want to use a bit of the _shouldSubmitForm if statement afterall - basically we need to use the callback if 
`(this._isTransactionFinished(data) || data.errorcode !== '0')`

pgs-gmartowski:
> Just trying this manually and need a couple of changes actually:
> 
> IE11 doesn't work with the callback you've put in (can it just be a simpler function?
Yes, it's just an example, I can simplify this Promise into sth else.
> 
> Also this gets fired on every response from the gateway instead of just the final result so we actually want to use a bit of the _shouldSubmitForm if statement afterall - basically we need to use the callback if
> `(this._isTransactionFinished(data) || data.errorcode !== '0')`
Ok, I'll change that.

PR #373, opened 2019-08-12
feature/ST-291-overwrite-translations

pgs-gmartowski:


src/ST.ts
```diff
@@ -251,6 +251,7 @@ class ST {
       this._threedinit = threedinit;
       this._cachetoken = cachetoken;
     }
+    localStorage.setItem('merchantTranslations', JSON.stringify(config.translations));
```
pgs-gmartowski: @james-simpson-ST here is my first concept (I think this simpler one) to store merchants translations in localStorage; question is if we can store such data in users browser - IMO it's not a sensitive data but it's better to ask before :)

The second one is to use WebWorker and IndexedDB (we cannot use ServiceWorker because of support for IE) - I'm currently working on this solution.

pgs-gmartowski:
@msakowicz I checked why automated tests are failing in this PR - there are translations in Polish specified in configuration object which override default translations; could you send me which translations I should leave as default to not make pipeline failed.

pgs-gmartowski:
> @msakowicz I checked why automated tests are failing in this PR - there are translations in Polish specified in configuration object which override default translations; could you send me which translations I should leave as default to not make pipeline failed.

Or we can resolve it in other way - you can change automated tests to recognize if there are some translations in configuration object. Whatever you want ;)

src/core/shared/Translator.ts
```diff
@@ -42,5 +43,12 @@ export class Translator {
...
   }
 
-  public translate = (text: string) => i18next.t(text);
+  /**
+   * Translates given text or gets it when it's defined by merchant in config.
+   * @param text
+   */
+  public translate = (text: string) => {
+    const translation = Utils.getLocalStorageItem(text, localStorage.merchantTranslations);
+    return translation ? translation : i18next.t(text, { content: text });
```
james-simpson-ST: Is there nothing in i18next that does this automatically i.e. if we pass it in as some kind of override into the i18next.init?

This works fine if not - I was just wondering if there was a way to re-use i18next more

pgs-gmartowski: I don't think there is any plugin which can resolve our problem in 100% - we need sth which can change existing, single translation (eg. 'An error occurred) and every translation in library as well depends on locale which we have.

PR #385, opened 2019-08-19
feature/ST-285-move-image.ts-base64-to-remote-server

pgs-gmartowski:


src/components/animated-card/animated-card-logos.ts
```diff
@@ -1,16 +1,28 @@
-import { amex, astropaycard, diners, discover, jcb, maestro, mastercard, piba, visa } from '../../core/imports/images';
+const filesToCache = {
```
pgs-gmartowski: info: I want to check if there is some other way to cache these resources instead of using webpack

PR #398, opened 2019-08-22
bugfix/ST-313-cannot-make-success-payment-if-previous-payment-was-declined

pgs-gmartowski:


pgs-gmartowski:
@james-simpson-ST I cannot approve my pull request, so I give 👍 here :)
Now I can fully focus on migrating animated card.

example/index.html
```diff
@@ -99,24 +99,13 @@ <h1 class="example-form__title">
...
-              'line-height-input': '12px'
+              'background-color-input': 'AliceBlue'
             },
             securityCode: {
-              'background-color-input': '#FFF333',
-              'background-color-input-error': '#f8d7da',
-              'color-input-error': '#3358FF',
-              'font-size-input': '20px',
-              'line-height-input': '20px'
+              'background-color-input': 'AliceBlue'
```
pgs-gmartowski: info: to get rid of the annoying test fields styling :/

src/core/classes/CommonFrames.class.ts
```diff
@@ -187,6 +190,7 @@ class CommonFrames extends RegisterFrames {
    */
   private _onTransactionComplete(data: any) {
     if ((this._isTransactionFinished(data) || data.errorcode !== '0') && this._submitCallback) {
+      this._validation.blockForm(false);
```
pgs-gmartowski: info: here is the place where must unblock form (we've got here fixes for both 313 and 316)

james-simpson-ST:
Could this be applied to the release branch instead of develop and then merged back? I think we will want to push this out faster so don't want to have to verify all the other tasks before we can release this

pgs-gmartowski:
> @pgs-gmartowski Could this be applied to the release branch instead of develop and then merged back? I think we will want to push this out faster so don't want to have to verify all the other tasks before we can release this

yup, sure, unfortunately we have an issue with translation feature here (automated tests are failing), perhaps I should revert these changes (which are currently in develop) and then we can make hotfix.

james-simpson-ST:
> > @pgs-gmartowski Could this be applied to the release branch instead of develop and then merged back? I think we will want to push this out faster so don't want to have to verify all the other tasks before we can release this
> 
> yup, sure, unfortunately we have an issue with translation feature here (automated tests are failing), perhaps I should revert these changes (which are currently in develop) and then we can make hotfix.

Yeah that makes sense, create a hotfix branch and we can then merge from hotfix into release/master and develop branches

PR #412, opened 2019-09-03
feature/ST-315-delete-animated-card-from-stjs-library

pgs-gmartowski:


pgs-gmartowski:
@james-simpson-ST I'm meriging this task, without it I cannot start connecting new version of animated  card

PR #445, opened 2019-09-23
feature/ST-328-add-update-jwt-functionality

pgs-gmartowski:


example/index.html
```diff
@@ -15,6 +15,19 @@
       <h1 class="example-form__title">
         Secure Trading<span>AMOUNT: <strong>10.00 GBP</strong></span>
       </h1>
+      <div class="example-form__section example-form__section--horizontal">
```
pgs-gmartowski: info: necessary for amount change event example

example/index.html
```diff
@@ -134,7 +147,8 @@ <h1 class="example-form__title">
           },
           translations: {
             'An error occurred': 'Wystąpił błąd'
-          }
+          },
+          updateJWT: true
```
pgs-gmartowski: info: only if we set this property we can use `updateJWT` method.

example/index.html
```diff
@@ -171,6 +185,10 @@ <h1 class="example-form__title">
             displayName: 'My Test Site'
           }
         });
+
+        document.getElementById('example-form-amount').addEventListener('input', function() {
+          st.updateJWT();
+        });
```
pgs-gmartowski: info: simple example of usage

pgs-gmartowski:
@james-simpson-ST not sure if this is sth what we wanted - updateJWT can be called anytime if `updateJWT` has been set to true (eg. when we set listener on AMOUNT field as it is in our example) page); otherwise it will call JSINIT in a standard way.

src/core/classes/StCodec.class.ts
```diff
@@ -108,6 +108,11 @@ class StCodec {
     }
   }
 
+  public static updateJWTValue(newJWT: string) {
+    StCodec.jwt = newJWT ? newJWT : StCodec.jwt;
```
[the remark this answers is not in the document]
pgs-gmartowski: true, changed

src/core/integrations/ApplePay.ts
```diff
@@ -242,7 +242,9 @@ export class ApplePay {
       : (this._buttonText = ApplePay.AVAILABLE_BUTTON_TEXTS[0]);
 
     // tslint:disable-next-line: max-line-length
-    this._applePayButtonProps.style = `-webkit-appearance: -apple-pay-button; -apple-pay-button-type: ${this._buttonText}; -apple-pay-button-style: ${this._buttonStyle}`;
+    this._applePayButtonProps.style = `-webkit-appearance: -apple-pay-button; -apple-pay-button-type: ${
```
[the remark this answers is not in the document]
pgs-gmartowski: good point, I think we must change `this._jwt` in ST class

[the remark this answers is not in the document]
pgs-gmartowski: and IMO that's it :)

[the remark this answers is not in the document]
james-simpson-ST: Wouldn't it have already passed this to the objects and created new copies of the variable inside of ApplePay/VisaCheckout themselves? And then inside of them they will have created a new Transport which in turn would have created a new Codec which is what actually needs updating.

Maybe we could use the messageBus in the codec to send the new JWT to all these places?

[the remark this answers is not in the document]
james-simpson-ST: I think the best approach might be to try out changing it in ST class and seeing if it works for VisaCheckout first (easiest one to test)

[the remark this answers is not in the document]
pgs-gmartowski: you're right, this is what I should've done first; I think communication via MessageBus is the best approach.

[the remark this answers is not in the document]
pgs-gmartowski: so now we should initialize VC and AP once again

example/index.html
```diff
@@ -109,7 +122,8 @@ <h1 class="example-form__title">
           },
           translations: {
             'An error occurred': 'Wystąpił błąd'
-          }
+          },
+          updateJWT: true // @TODO change the name of variable
```
james-simpson-ST: I think this should be called deferInit as this might be used for other purposes in the future if we need to defer anything and doesn't have to be restricted by just being used for the updateJWT functionality

src/ST.ts
```diff
@@ -326,6 +345,14 @@ class ST {
     return new visa(config, this._jwt, this._gatewayUrl);
   }
 
+  public updateJWT(newJWT: string) {
+    this._jwt = newJWT;
```
james-simpson-ST: Should this throw an error if the merchant hasn't set _updateJWT to true as it wont work correctly?

src/components/control-frame/ControlFrame.ts
```diff
@@ -151,6 +156,10 @@ class ControlFrame extends Frame {
     StCodec.jwt = StCodec.originalJwt;
   }
 
+  private _onUpdateJWT(jwt: string) {
+    StCodec.jwt = jwt;
```
james-simpson-ST: I believe here we also need to update originalJwt. There is a bug at the moment that if a transaction is processed but declines (using 4242 4242 4242 4242 card) and then another payment with 4111 1111 1111 1111 card is processed, the 2nd transaction will process with the original JWT instead of the updated JWT.

james-simpson-ST: I tried a quick change of this and it seemed to work by just adding extra

`StCodec.originalJwt = jwt;`

src/components/control-frame/ControlFrame.ts
```diff
@@ -151,6 +156,10 @@ class ControlFrame extends Frame {
...
+  }
```
james-simpson-ST: We should have unit tests for this and actually need unit tests to cover all changes e.g. new updateJWT function in ST.ts etc

example/index.html
pgs-gmartowski: true, done

src/ST.ts
pgs-gmartowski: good point ! 
changed

src/components/control-frame/ControlFrame.ts
pgs-gmartowski: oops, yes you're right. 
Done.

src/components/control-frame/ControlFrame.ts
pgs-gmartowski: I've added couple of UT; there are also tests for this class in ST-318

PR #446, opened 2019-09-24
Feature/st 346 ensure styles are backwards compatible

james-simpson-ST:


src/core/classes/RegisterFrames.class.ts
```diff
@@ -61,4 +61,19 @@ export default class RegisterFrames {
...
     return [];
   }
+
+  /**
+   * Set defaultStyles if styles are defined as a flat structure
+   * @param styles
+   */
+  private _getStyles(styles: any) {
+    // If we have an object as a value then we assume we are in the new format otherwise we are a flat array
+    for (const key in styles) {
```
pgs-gmartowski: not sure if there should be `// tslint:disable-next-line:forin` before `for` loop, due to tslint settings

james-simpson-ST: It's only needed if theres not an if statement straight after the for loop 
https://palantir.github.io/tslint/rules/forin/

In this case we have if statement on next line so it doesn't complain

PR #479, opened 2019-10-10
feature/ST-352-add-posibility-to-use-config-from-mock-json

pgs-gmartowski:


example/index.html
```diff
@@ -81,71 +81,80 @@ <h1 class="example-form__title">
...
-          'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJhbTAzMTAuYXV0b2FwaSIsImlhdCI6MTU2MDk1NDA4OC45Njc5MDM0LCJwYXlsb2FkIjp7ImJhc2VhbW91bnQiOiIxMDAwIiwiYWNjb3VudHR5cGVkZXNjcmlwdGlvbiI6IkVDT00iLCJjdXJyZW5jeWlzbzNhIjoiR0JQIiwic2l0ZXJlZmVyZW5jZSI6InRlc3RfamFtZXMzODY0MSIsImxvY2FsZSI6ImVuX0dCIn19.ehveog4Xds-ak2yOhSgzZd-yzu4JZoC-a3xKqY_HJOU';
-
-        var jwt = parsedUrl.searchParams.get('jwt') || cardinalJwt;
-
-        var st = SecureTrading({
-          jwt: jwt, // Can't use property shorthand because it isn't supported by IE
-          styles: {
-            defaultStyles: {
-              'background-color-input': 'AliceBlue'
+        var ST = SecureTrading;
```
james-simpson-ST: Now that index.html can be used for all configs do we still need bypass.html and immediate.html or can these be removed?

example/index.html
```diff
@@ -81,71 +81,80 @@ <h1 class="example-form__title">
...
+          const animatedCardConfig = data.animatedCardConfig;
+          const submitOnSuccessConfig = data.submitOnSuccessConfig;
+          const translationsConfig = data.translationsConfig;
+          const componentsConfig = data.componentsConfig;
+          const applePayConfig = data.applePayConfig;
+          const visaCheckoutConfig = data.visaCheckoutConfig;
+          var parsedUrl = new URL(window.location.href);
+          var cardinalJwt = cardinalJwtConfig
+            ? cardinalJwtConfig
+            : 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJhbTAzMTAuYXV0b2FwaSIsImlhdCI6MTU2NjI5NjUwOS45MDMwODU1LCJwYXlsb2FkIjp7ImJhc2VhbW91bnQiOiIxMDAwIiwiYWNjb3VudHR5cGVkZXNjcmlwdGlvbiI6IkVDT00iLCJjdXJyZW5jeWlzbzNhIjoiR0JQIiwic2l0ZXJlZmVyZW5jZSI6InRlc3RfamFtZXMzODY0MSIsImxvY2FsZSI6ImVuX0dCIn19.8vDjHQB98E2U2bJtOu4t61YKyLc964cLdmpvX8R9L8w';
```
james-simpson-ST: Should we simplify this and not have defaults in index.html, instead just have all config in config.json so we don't have lots of places containing large JWT strings?

example/index.html
```diff
@@ -81,71 +81,80 @@ <h1 class="example-form__title">
...
+          var applePayVisaCheckoutJwt = applePayVisaCheckoutJwtConfig
+            ? applePayVisaCheckoutJwtConfig
+            : 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJhbTAzMTAuYXV0b2FwaSIsImlhdCI6MTU2MDk1NDA4OC45Njc5MDM0LCJwYXlsb2FkIjp7ImJhc2VhbW91bnQiOiIxMDAwIiwiYWNjb3VudHR5cGVkZXNjcmlwdGlvbiI6IkVDT00iLCJjdXJyZW5jeWlzbzNhIjoiR0JQIiwic2l0ZXJlZmVyZW5jZSI6InRlc3RfamFtZXMzODY0MSIsImxvY2FsZSI6ImVuX0dCIn19.ehveog4Xds-ak2yOhSgzZd-yzu4JZoC-a3xKqY_HJOU';
+          var jwt = parsedUrl.searchParams.get('jwt') || cardinalJwt;
+          var st = ST({
```
james-simpson-ST: Should we just have one stConfig instead of all these being separate, then just update stConfig with submitCallback in here?

That way if we add a new config option this won't need to change to be able to test the additional config option

james-simpson-ST: If you agree please work with Marek to update the config provided back in the tests

example/index.html
pgs-gmartowski: I'm for it, I'll ask Marek if we really need this in index.html

example/index.html
pgs-gmartowski: yes :) I forgot about it

PR #480, opened 2019-10-14
feature/ST-356-add-option-to-use-stage/prod-urls-like-in-APMs

pgs-gmartowski:


example/index.html
```diff
@@ -101,6 +101,7 @@ <h1 class="example-form__title">
           },
           analytics: true,
           animatedCard: true,
+          livestatus: 0,
```
pgs-gmartowski: @james-simpson-ST 
1. Could livestatus be here in the main configuration (It will be common for all payment types) ?
2. Shouldn't it be boolean ?

src/ST.ts
```diff
@@ -263,6 +264,7 @@ class ST {
       this._threedinit = threedinit;
       this._cachetoken = cachetoken;
     }
+    this._livestatus = livestatus;
```
james-simpson-ST: If live status is not defined we want to default to 0 (testing) so merchants don't accidentally put their transactions live

pgs-gmartowski: Good point !

pgs-gmartowski: Done; additionally we have `livestatus` in interface `IConfig` set as obligatory.

james-simpson-ST: This should be changed to optional as I don't want to break merchants currently testing that have not passed in livestatus while testing

pgs-gmartowski: Oh, I see, ok sure

pgs-gmartowski: ok, done

PR #513, opened 2019-10-31
develop_paymentcardrefactor

pgs-gmartowski:


pgs-gmartowski:
@james-simpson-ST I'll create new branch for unit tests (couple of classes lost its code coverage after my refactor).

james-simpson-ST:
I was expecting to see that BinLookup could also be removed from the js-payments so we're not duplicating that logic between the two libraries. Maybe we need to expose the binLookup call as a function in the js-payments-card so that it can be called and any required attributes can be accessed by the js-payments library -  do we want a task for this?

pgs-gmartowski:
> I was expecting to see that BinLookup could also be removed from the js-payments so we're not duplicating that logic between the two libraries. Maybe we need to expose the binLookup call as a function in the js-payments-card so that it can be called and any required attributes can be accessed by the js-payments library - do we want a task for this?

Good point, I think it might be moved to the separate package which can be imported. 
Yes, I'll add task for this tomorrow.
I think we can also delete couple of dependencies in js-payments and js-payments-card as well.

src/json/credentials.json
```diff
@@ -0,0 +1,9 @@
+{
+  "jwt": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJhbTAzMTAuYXV0b2FwaSIsImlhdCI6MTU3Nzk2ODQwMC4wODgzNDg2LCJwYXlsb2FkIjp7Im1haW5hbW91bnQiOiIxMC4wMCIsImFjY291bnR0eXBlZGVzY3JpcHRpb24iOiJFQ09NIiwiY3VycmVuY3lpc28zYSI6IkdCUCIsInNpdGVyZWZlcmVuY2UiOiJ0ZXN0X2phbWVzMzg2NDEiLCJsb2NhbGUiOiJlbl9HQiJ9fQ.TIEvASL1KHqYlQ0j_kadK0x9JFXlZ13W8jEyEEMQQWo",
+  "visaCheckout": {
+    "merchantId": "DUT1MEXJO10RARJF2S521ImTyKfn3_JmxePdXcydQIUb4kx4"
+  },
+  "applePay": {
+    "merchantId": "merchant.net.securetrading.test"
+  }
+}
```
pgs-gmartowski: @james-simpson-ST, @wirus15 - this is not exactly what Maciej mentioned in comment, but IMO we should somehow provide 'testing credentials' for users. I'm not sure if it should be here, but it has to be provided some way. I think the other options are:
- attaching credentials in markdown file (description of library)
- sending credentials to merchants by email ('on demand')
Guys, what do you think `bout this ?

james-simpson-ST: I agree we need to better direct our merchants to signup. We've recently improved the docs for the JS so we probably want to change the link to "API documentation" to point to https://docs.securetrading.com/document/javascript-library/ and maybe make it stand out more that they should go there to find out how to use the library - maybe move it to the top instead of the Quickstart section?

james-simpson-ST: In terms of merchants getting the credentials, the documents outline how to do this

PR #528, opened 2019-11-07
feature/ST-385-link-js-payments-card-to-js-payments-library

pgs-gmartowski:


example/index.ts
```diff
@@ -4,3 +4,10 @@
  * It can be treated as a reference for merchants how to integrate with STJS.
  */
 import './style.scss';
+import Card from '@securetrading/js-payments-card/dist/stcard.js';
+import '@securetrading/js-payments-card/dist/stcardstyle.css';
```
pgs-gmartowski: info: this is just for now; in new version of js-payments-card we've got new path without `/dist/`

PR #530, opened 2019-11-12
Hotfix/apple pay supported networks

james-simpson-ST:
We now use the supportedNetworks passed in by the merchant in the config to determine a subset of supportedNetworks based on what is supported by the device

### DIFF
--- a/src/core/integrations/ApplePay.ts
+++ b/src/core/integrations/ApplePay.ts
@@ -231,15 +231,28 @@ export class ApplePay {
    * @private
    */
   private _setSupportedNetworks() {
+    let supportedNetworks;
     if (this.applePayVersion <= ApplePay.APPLE_PAY_MIN_VERSION) {
-      this._paymentRequest.supportedNetworks = ApplePay.BASIC_SUPPORTED_NETWORKS;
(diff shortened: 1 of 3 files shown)


src/core/integrations/ApplePay.ts
```diff
@@ -231,15 +231,28 @@ export class ApplePay {
...
     ) {
-      this._paymentRequest.supportedNetworks = ApplePay.VERSION_4_SUPPORTED_NETWORKS;
+      supportedNetworks = ApplePay.VERSION_4_SUPPORTED_NETWORKS;
     } else {
-      this._paymentRequest.supportedNetworks = ApplePay.VERSION_5_SUPPORTED_NETWORKS;
+      supportedNetworks = ApplePay.VERSION_5_SUPPORTED_NETWORKS;
+    }
+    if (this._paymentRequest.supportedNetworks.length > 0) {
+      const userDefinedSupportedNetworks: string[] = [];
+      let i;
```
pgs-gmartowski: should this be outside `for ... in loop` ? why not  `for (let i in supportedNetworks) {`

james-simpson-ST: Updated to put into for loop :)

PR #531, opened 2019-11-13
Only process JSINIT at this point if deferInit is true

james-simpson-ST:
Fixes STJS-389


pgs-gmartowski:
> Fixes STJS-389

Good point !

james-simpson-ST:
I think this needs further testing and maybe a different fix. I have just tried to use deferInit and updateJWT and this doesn't seem to be working @msakowicz could you set up some test cases for this and then @pgs-gmartowski could you take over this bugfix to investigate why the amount is not changing to the correct amount. It seems as though the init is being fired twice even if deferInit is true

We need to test 
1) deferInit: true and updateJWT works as intended (i.e. uses the new amount after the JWT is updated)
2) startOnLoad works when deferInit: false (i.e. the JWT contains the pan/expiry/securitycode and the payment is performed straight away)

PR #532, opened 2019-11-13
feature/ST-379-migrate-formatting-and-validation

pgs-gmartowski:


src/components/control-frame/ControlFrame.ts
```diff
@@ -396,14 +396,18 @@ class ControlFrame extends Frame {
       this._formFields,
       deferInit
     );
+    console.error(validity);
```
james-simpson-ST: We should remove these console logs

pgs-gmartowski: oops, sorry :)

pgs-gmartowski: done

PR #538, opened 2019-11-13
feature/ST-377-hide-cvv-field-continuation-in-js-payments

pgs-gmartowski:


pgs-gmartowski:
Still couple of unit tests needed here, but generally it can be reviewed to make sure if this is what we want to have

pgs-gmartowski:
@james-simpson-ST I've made 'small' refactor here too ;)

PR #585, opened 2019-12-09
Hotfix/serialize javascript vulnerability

pgs-gmartowski:


pgs-gmartowski:
@james-simpson-ST There is a one plugin in which we still got obsolete serialize-javascript library version, it's copy-webpack-plugin - we should raise new issue in it's repo - they've got a bug in code.

PR #586, opened 2019-12-09
feature/ST-443-provide-translated-labels-to-js-payments-card

pgs-gmartowski:


src/components/animated-card/animated-card.ts
```diff
@@ -8,10 +8,10 @@ if (Card && document.URL.includes('animated')) {
   // @ts-ignore
   const card: Card = new Card({
     animatedCardContainer: 'st-animated-card',
-    locale: 'en_GB'
+    locale: localStorage.getItem('locale') ? localStorage.getItem('locale') : 'en_GB'
```
pgs-gmartowski: info: it looks like we can use localStorage to provide current language and we don't need to change anything in js-payments-card

PR #603, opened 2019-12-16
feature/ST-400-control-frame-ut

pgs-gmartowski:


src/components/control-frame/ControlFrame.ts
```diff
@@ -3,6 +3,7 @@ import { StCodec } from '../../core/classes/StCodec.class';
 import {
   FormFieldsDetails,
   FormFieldsValidity,
+  IDecodedJwt,
```
pgs-gmartowski: info: one more `interface` needed - BTW there are more places where we have `any` and we have to get rid of them as soon as it's possible.

src/core/models/ControlFrame.ts
```diff
@@ -58,11 +58,28 @@ const FormFieldsValidity: IFormFieldsValidity = {
...
+  payload: {
+    accounttypedescription: string;
+    baseamount: string;
+    currencyiso3a: string;
+    locale: string;
+    mainamount: string;
+    pan: string;
+    expirydate: string;
+    securitycode: string;
+    sitereference: string;
```
james-simpson-ST: Remind me, if a merchant included an additional field in there would this raise an error? 

Merchants can add other fields that they want to be passed onto the ST backend inside the JWT payload. Our library should just pass this on for the merchant. So as long as it just ignores any un-recognised field that is fine but if it raised an error if the field is un-recognised we'd have to allow a wildcard field

src/components/control-frame/ControlFrame.ts
james-simpson-ST: Agreed, we should be more strict whenever we notice we're still using "any"

src/core/models/ControlFrame.ts
pgs-gmartowski: yes, it'll be treated as a wrong property. I just wanted to reduce `any` to minimum, but if we have a case like above, yes we must allow merchants to provide sth which is not included in the interface.

wirus15: Interfaces are only checked during compilation so any other parameters in jwt payload will be passed without errors

src/components/control-frame/ControlFrame.ts
```diff
@@ -296,8 +298,8 @@ class ControlFrame extends Frame {
   }
 
   private _getPan() {
-    const decodedJwt: any = JwtDecode<IStJwtObj>(this.params.jwt);
-    return decodedJwt.payload.pan ? JwtDecode<IStJwtObj>(this.params.jwt).payload.pan : this._cardNumber;
+    this._decodedJwt = JwtDecode(this.params.jwt);
```
wirus15: Unless you hava a good reason changing object's state inside a getter is a bad idea

src/core/models/ControlFrame.ts
james-simpson-ST: In that case this won't be a problem, just wanted to make sure I was remembering correctly

src/components/control-frame/ControlFrame.ts
pgs-gmartowski: true, I changed this to return only 'pan' and get rid of one extra method

PR #613, opened 2019-12-18
fix/ST-357-card-frames-fix

pgs-gmartowski:


src/ST.ts
```diff
@@ -4,7 +4,7 @@ import 'location-origin';
 import { debounce } from 'lodash';
 import 'url-polyfill';
 import 'whatwg-fetch';
-import CardFrames from './core/classes/CardFrames.class';
+import { CardFrames } from './core/classes/CardFrames.class';
```
pgs-gmartowski: @wirus15 according to our yesterdays discussion, we'll slowly get rid of default import / exports

src/core/classes/CardFrames.class.ts
```diff
@@ -77,7 +77,7 @@ class CardFrames extends RegisterFrames {
     this.onInit();
   }
 
-  protected onInit() {
+  protected onInit(): void {
```
pgs-gmartowski: same here; if the method returns nothing, in fact it returns void

src/core/classes/CardFrames.class.ts
```diff
@@ -116,18 +116,25 @@ class CardFrames extends RegisterFrames {
...
+      fields.length >= CardFrames.COMPLETE_FORM_NUMBER_OF_FIELDS &&
+      targets.length >= CardFrames.COMPLETE_FORM_NUMBER_OF_FIELDS &&
+      fields.length === targets.length
+    ) {
       targets.map((item, index) => {
-        document.getElementById(item).appendChild(fields[index]);
+        const element: HTMLElement = document.getElementById(item);
+        if (element) {
+          element.appendChild(fields[index]);
+        }
```
pgs-gmartowski: here we have some kind of protection when merchant doesn't specify fields correctly

wirus15: maybe we should raise an error with clear message when someone wants to bind the field to the target that doesn't exist? right now it could be hard to debug why given field is not showing

src/core/classes/CardFrames.class.ts
```diff
@@ -140,7 +147,7 @@ class CardFrames extends RegisterFrames {
     } else if (this._noFieldConfiguration) {
       return [];
     } else {
-      return false;
+      return [this.componentIds.cardNumber, this.componentIds.expirationDate, this.componentIds.securityCode];
```
wirus15: this would look better if split to separate lines

src/core/classes/CardFrames.class.ts
```diff
@@ -388,4 +397,4 @@ class CardFrames extends RegisterFrames {
   }
 }
 
-export default CardFrames;
+export { CardFrames };
```
wirus15: imo it's more convenient to put `export` next to the class name `export class CardFrames { ... }`

src/json/config.json
```diff
@@ -1,10 +1,10 @@
 {
   "analytics": true,
-  "animatedCard": true,
+  "animatedCard": false,
   "deferInit": false,
   "fieldsToSubmit": ["pan", "expirydate", "securitycode"],
   "buttonId": "merchant-submit-button",
-  "jwt": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJhbTAzMTAuYXV0b2FwaSIsImlhdCI6MTU3NjU5MTM0OS4xNzIyODI3LCJwYXlsb2FkIjp7Im1haW5hbW91bnQiOiIxMC4wMCIsImFjY291bnR0eXBlZGVzY3JpcHRpb24iOiJFQ09NIiwiY3VycmVuY3lpc28zYSI6IkdCUCIsInNpdGVyZWZlcmVuY2UiOiJ0ZXN0X2phbWVzMzg2NDEiLCJsb2NhbGUiOiJlbl9HQiJ9fQ.X-fYUQTU3JY2LE6eWA4nJ56s-GaLm_vR7-ULaL3y08M",
+  "jwt": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJhbTAzMTAuYXV0b2FwaSIsImlhdCI6MTU3NjY2MTUyNi4zMjU2MzE0LCJwYXlsb2FkIjp7Im1haW5hbW91bnQiOiIxMC4wMCIsImFjY291bnR0eXBlZGVzY3JpcHRpb24iOiJFQ09NIiwiY3VycmVuY3lpc28zYSI6IkdCUCIsInNpdGVyZWZlcmVuY2UiOiJ0ZXN0X2phbWVzMzg2NDEiLCJsb2NhbGUiOiJlbl9HQiJ9fQ.NHUBwU3sjvP1iFDNwnyXhZy-_74Uh0X3Xw86IfQRCjI",
```
wirus15: Generally this file shouldn't be inside the repo at all. It should be listed in `.gitignore` and we should have another file called `config.json.dist` with default configuration for most of the fields and fields like the `jwt` should be empty

src/core/classes/CardFrames.class.ts
pgs-gmartowski: yes, changed; I also forgot to check if the animated card is set as available or not - also fixed

src/core/classes/CardFrames.class.ts
```diff
@@ -116,18 +116,25 @@ class CardFrames extends RegisterFrames {
...
    */
-  protected registerElements(fields: HTMLElement[], targets: string[]) {
-    if (fields.length && targets.length) {
+  protected registerElements(fields: HTMLElement[], targets: string[]): void {
+    if (
+      fields.length >= CardFrames.COMPLETE_FORM_NUMBER_OF_FIELDS &&
+      targets.length >= CardFrames.COMPLETE_FORM_NUMBER_OF_FIELDS &&
+      fields.length === targets.length
+    ) {
       targets.map((item, index) => {
```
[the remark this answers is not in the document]
pgs-gmartowski: true, there is no need to return any array here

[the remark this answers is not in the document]
pgs-gmartowski: fixed

src/core/classes/CardFrames.class.ts
pgs-gmartowski: actually prettier automatically formats it to one liner :/ I've found workaround for this - https://github.com/prettier/prettier-vscode/issues/352.
Looks like a know problem :(

src/core/classes/CardFrames.class.ts
pgs-gmartowski: fixed

src/json/config.json
pgs-gmartowski: ok, but where should this file be placed ? If you didn't specify jwt in config, library'll throw an error and prevent for making test payments,

src/core/classes/CardFrames.class.ts
```diff
@@ -116,18 +116,25 @@ class CardFrames extends RegisterFrames {
    * @param fields
    * @param targets
    */
-  protected registerElements(fields: HTMLElement[], targets: string[]) {
-    if (fields.length && targets.length) {
+  protected registerElements(fields: HTMLElement[], targets: string[]): void {
```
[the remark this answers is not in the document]
pgs-gmartowski: good point, but I suppose it should be changed somewhere in ST core class - could we extract this to another task ?

pgs-gmartowski:
@wirus15 I've added comments and fixes.

src/core/classes/CardFrames.class.ts
wirus15: All walkarounds from this issue are terrible except for dropping prettier in favor of tslint 😁

src/json/config.json
wirus15: It should be in the same place as the current file. It's up to the developer who downloads the source code to make a copy of this file, name it `config.json` and fill the required properties like jwt.

src/core/classes/CardFrames.class.ts
```diff
@@ -105,35 +100,30 @@ class CardFrames extends RegisterFrames {
...
       targets.length >= CardFrames.COMPLETE_FORM_NUMBER_OF_FIELDS &&
       fields.length === targets.length
     ) {
-      targets.map((item, index) => {
+      targets.forEach((item, index) => {
         const element: HTMLElement = document.getElementById(item);
-        if (element) {
+        try {
           element.appendChild(fields[index]);
+        } catch (e) {
```
wirus15: This changed nothing actually. What I meant was to throw an error with a clear message what went wrong, something like `A target element for the input field with id ... could not be found. Please check your configuration`

src/core/classes/CardFrames.class.ts
```diff
@@ -147,16 +137,14 @@ class CardFrames extends RegisterFrames {
     } else if (this._noFieldConfiguration) {
       return [];
     } else {
-      return [this.componentIds.cardNumber, this.componentIds.expirationDate, this.componentIds.securityCode];
+      return [
+        this.componentIds.cardNumber, //
```
wirus15: still don't like it but it looks like it's the only sollution for now

src/core/classes/CardFrames.class.ts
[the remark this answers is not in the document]
pgs-gmartowski: Added to backlog

src/core/classes/CardFrames.class.ts
pgs-gmartowski: oh, now I see ..
I've added this as translation (we need it in couple more languages then).
I also added `@babel/plugin-proposal-optional-catch-binding` to omit `catch` parameter which is not specified in our case.

src/json/config.json
pgs-gmartowski: ok, fixed; I'm not sure if it should look like this; could you check it ?

PR #654, opened 2020-01-10
feature/ST-422-change-library-target-to-umd

pgs-gmartowski:


example/index.html
```diff
@@ -91,63 +91,64 @@ <h1 class="example-form__title">
...
-            if (data) {
-              configurationInit(ST, data);
-            } else {
-              fetchDefaultConfig();
-            }
-          })
-          .catch(function() {
-            fetchDefaultConfig();
-          });
+    <!--    <script type="text/javascript" src="st.js"></script>-->
```
james-simpson-ST: Have we proven that using the library this way still works? As in they can do it without using npm and just including the script src tag?

pgs-gmartowski: yes, it's working as an import and as on script as well.
If merchant wanted to launch the example page with imported version of st.js, he has to delete comments in init.ts file.

james-simpson-ST: Okay great thanks for clarification

PR #702, opened 2020-01-21
STJS-424 example page

wirus15:
I replaced the example page with the already compiled version and removed it from the webpack config.

I also added the example page to the docker container. It's being served by the same nginx server but under different port. It was a simpler solution because I didn't want to create a separate image for it.


pgs-gmartowski:
> I replaced the example page with the already compiled version and removed it from the webpack config.
> 
> I also added the example page to the docker container. It's being served by the same nginx server but under different port. It was a simpler solution because I didn't want to create a separate image for it.

What should we do if we want to test application locally ? I mean, if the example page is is still available on localhost ?

james-simpson-ST:
The concept looks fine but it seems as though it's failing to run npm run prod in the pipeline?

wirus15:
@james-simpson-ST I updated the configs and the build passes now. There is however a second build that I don't understand, with older `travis.yml` config and this one naturally fails. What are those two builds?

james-simpson-ST:
> @james-simpson-ST I updated the configs and the build passes now. There is however a second build that I don't understand, with older `travis.yml` config and this one naturally fails. What are those two builds?

One is the PR and one is the Push, however, I was expecting both to use the same Travis config, not sure why one would work and one would fail - looking at the latest builds after your latest merge from docker-refactor it looks like it's both using the same config to me?

wirus15:
Yeah, it seems that it fixed itself :)

PR #713, opened 2020-01-24
STJS-404 added mocked cybertonica gateway

wirus15:


src/core/classes/StTransport.class.ts
```diff
@@ -46,14 +46,12 @@ class StTransport {
    * @return A Promise object that resolves the gateway response
    */
   public async sendRequest(requestObject: IStRequest) {
-    console.error({ ...StTransport.DEFAULT_FETCH_OPTIONS, body: this._codec.encode(requestObject) });
```
pgs-gmartowski: thx ;)

src/core/integrations/Cybertonica.ts
```diff
@@ -1,41 +1,43 @@
 import { environment } from '../../environments/environment';
-import StTransport from '../classes/StTransport.class';
 import {
   IAFCybertonica,
   ICybertonicaInitQuery,
-  ICybertonicaPostQuery,
-  ICybertonicaPostResponse
+  ICybertonicaPostQuery, ICybertonicaPostResponse,
+  ICybertonicaPostResponseStatus
```
pgs-gmartowski: this could be moved to `/models` (generally Cybertonica model can be splitted)

src/core/classes/StTransport.class.ts
wirus15: no problem 😁

pgs-gmartowski:
👍

src/core/integrations/cybertonica/MockedCybertonicaGateway.ts
```diff
@@ -0,0 +1,35 @@
...
+  /** @todo: remove static after implementing DI and singleton services */
+  private static responseStatus: ICybertonicaPostResponseStatus;
+
+  /** @todo: remove static after implementing DI and singleton services */
+  static setResponseStatus(responseStatus: ICybertonicaPostResponseStatus): void {
+    MockedCybertonicaGateway.responseStatus = responseStatus;
+  }
+
+  postQuery(query: ICybertonicaPostQuery): Promise<ICybertonicaPostResponse> {
+    return Promise.resolve({
```
james-simpson-ST: For the other mocks we perform a fetch request to the wiremock server, this allows us to return different responses for different test cases e.g. fraudcontrolshieldstatuscode ALLOW/CHALLENGE/DENY, do we not need the same here so that Marek can add a test case for each of the shield status codes that can be returned as the code will behave differently depending on the status.

wirus15: I guess you're right, this is actually not needed. We use wiremock in black box testing, and for unit tests we can simply mock that method. This just one of my habbits from PHP where you can't change the behavior of existing classes just like that :)

PR #718, opened 2020-01-26
Feature/st 460 use imported binlookup

james-simpson-ST:


src/components/card-number/CardNumber.ts
```diff
@@ -1,13 +1,14 @@
...
-import { BinLookup } from '../../core/shared/BinLookup';
 import { Formatter } from '../../core/shared/Formatter';
 import { FormField } from '../../core/shared/FormField';
 import { Language } from '../../core/shared/Language';
 import { MessageBus } from '../../core/shared/MessageBus';
 import { Selectors } from '../../core/shared/Selectors';
 import { Utils } from '../../core/shared/Utils';
 import { Validation } from '../../core/shared/Validation';
+// @ts-ignore
+import { PaymentsUtils } from '@securetrading/js-payments-utils';
```
pgs-gmartowski: I wonder if this should be named as `utils`, maybe a better way is to name this module 'BinLookup` ?
It will be coherent with SOLID principle.

pgs-gmartowski: @wirus15 what do you think about this ?

james-simpson-ST: The package also includes the LuhnCheck so not sure if BinLookup/IinLookup is the best name, but I agree utils isn't the best name either. I'm just not sure what is the best name here so any suggestions are welcome

src/core/classes/RegisterFrames.class.ts
```diff
@@ -1,7 +1,8 @@
 import { IStyles } from '../models/IStyles';
-import { BinLookup } from '../shared/BinLookup';
 import { MessageBus } from '../shared/MessageBus';
 import { StJwt } from '../shared/StJwt';
+// @ts-ignore
+import { PaymentsUtils } from '@securetrading/js-payments-utils';
```
pgs-gmartowski: IMO this is not needed anymore

james-simpson-ST: Yeah you're right, I've removed it 😄

PR #720, opened 2020-01-28
bugfix/ST-403-duplicated-songbirdjs

pgs-gmartowski:


src/core/integrations/CardinalCommerce.ts
```diff
@@ -155,7 +155,7 @@ export class CardinalCommerce {
   }
 
   protected _threeDSetup() {
-    if (!this._called) {
+    if (!this._called && !DomMethods.isScriptLoaded(this._sdkAddress)) {
```
james-simpson-ST: Should it be the responsibility of the insertScript to check isScriptLoaded? That way we don't need to change both CardinalCommerce and VisaCheckout. If it isn't done this way don't we need to change VisaCheckout?

Also do we still need to call _onCardinalLoad if we have a fresh instance of CardinalCommerce - currently I don't think we will but I just wanted to check we don't want to?

src/json/config.json
```diff
@@ -29,7 +29,7 @@
       "color-error": "#3358FF"
     }
   },
-  "submitOnSuccess": false,
+  "submitOnSuccess": true,
```
james-simpson-ST: Does this want to be reverted before we merge back?

james-simpson-ST:
Only just noticed this is a WIP - but I've done the review now anyway 😉

src/core/integrations/CardinalCommerce.ts
pgs-gmartowski: Ad.1. Yes, your're right I'll fix this.
Ad.2. Not sure, bu probably this is not needed; I'll take a look at this.

src/json/config.json
pgs-gmartowski: yup, thx, this is one of my `debugging` changes ;)

src/core/integrations/CardinalCommerce.ts
pgs-gmartowski: Ad.1. Done,
Ad.2. Seems to work fine.

src/json/config.json
pgs-gmartowski: done

src/core/shared/DomMethods.ts
```diff
@@ -1,106 +1,118 @@
...
+    return script;
   }
 
-  public static addListener(targetId: string, listenerType: string, callback: any) {
-    document.getElementById(targetId).addEventListener(listenerType, callback);
+  public static insertStyle(contents: string): void {
+    const head = document.getElementById('insertedStyles');
+    if (head) {
+      return;
+    }
```
pgs-gmartowski: info: this change has been added because of duplicating styles in head whih was inserted by updateJwt function

PR #736, opened 2020-02-04
Bugfix/st 453 block form after payment

james-simpson-ST:
Additional functionality changes to prevent customers processing duplicate payments:
1) Form will now be permanently blocked after successful payment (can't process two transactions with the same payment form)
2) Form button will remain disabled after successful payment but display "Pay"
3) Success message will remain indefinitely (all other messages will keep current behaviour of disappearing after 5seconds)

### DIFF
--- a/src/components/control-frame/ControlFrame.ts
+++ b/src/components/control-frame/ControlFrame.ts
@@ -2,7 +2,7 @@ import JwtDecode from 'jwt-decode';
 import { StCodec } from '../../core/classes/StCodec.class';
 import { FormFieldsDetails } from '../../core/models/constants/FormFieldsDetails';
 import { FormFieldsValidity } from '../../core/models/constants/FormFieldsValidity';
-import { FormState } from "../../core/models/constants/FormState";
+import { FormState } from '../../core/models/constants/FormState';
 import { ICard } from '../../core/models/ICard';
 import { IDecodedJwt } from '../../core/models/IDecodedJwt';
 import { IFormFieldsDetails } from '../../core/models/IFormFieldsDetails';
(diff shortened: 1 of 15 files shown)


src/components/control-frame/ControlFrame.ts
```diff
@@ -229,12 +230,14 @@ export class ControlFrame extends Frame {
...
         this._notification.success(Language.translations.PAYMENT_SUCCESS);
+        this._validation.blockForm(FormState.COMPLETE);
       })
       .catch((error: any) => {
         this._notification.error(Language.translations.PAYMENT_ERROR);
+        this._validation.blockForm(FormState.AVAILABLE);
       })
       .finally(() => {
-        this._validation.blockForm(false);
+        ControlFrame._resetJwt();
```
pgs-gmartowski: info: so right now we need to refresh page to make payment once again ?

src/components/expiration-date/ExpirationDate.ts
```diff
@@ -29,8 +30,8 @@ export class ExpirationDate extends FormField {
   }
 
   public setDisableListener() {
-    this.messageBus.subscribe(MessageBus.EVENTS.BLOCK_EXPIRATION_DATE, (state: boolean) => {
-      state ? this._disableInputField() : this._enableInputField();
+    this.messageBus.subscribe(MessageBus.EVENTS.BLOCK_EXPIRATION_DATE, (state: FormState) => {
+      state!==FormState.AVAILABLE ? this._disableInputField() : this._enableInputField();
```
pgs-gmartowski: sth happens with formatting (no space between state and status)

src/components/control-frame/ControlFrame.ts
james-simpson-ST: Yes, one payment per page load, I can't think of a use case where multiple payments would be made on the same checkout. If a merchant requires this in the future I think we can allow it via the config but I think this is the better default behaviour

src/components/expiration-date/ExpirationDate.ts
james-simpson-ST: Fixed

src/components/control-frame/ControlFrame.ts
wirus15: I think this might be the case in SPAs. There a client could make multiple payments for different stuff without fully reloading the page.

PR #741, opened 2020-02-07
Fixes for formId and JWT containing payment data

james-simpson-ST:


src/ST.ts
```diff
@@ -100,18 +104,18 @@ class ST {
 
   private init(config: IConfig): void {
     this._config = this._configuration.init(config);
+    // TODO theres probably a better way rather than having to remember to update Selectors
+    Selectors.MERCHANT_FORM_SELECTOR = this._config.formId;
```
pgs-gmartowski: I agree, but we can do this after fixing bugs (or I can change this today if I'll have a moment)

PR #753, opened 2020-02-12
[WIP] bugfix/ST-528-allowed-requests-list

pgs-gmartowski:


pgs-gmartowski:
@james-simpson-ST, @wirus15 I think it should be called `hotfix` branch, or we can add this to the next release.

PR #755, opened 2020-02-13
[WIP] hotfix/animated-card-style-issue

pgs-gmartowski:


src/components/animated-card/animated-card.scss
```diff
@@ -0,0 +1,8 @@
+.st-animated-card-iframe {
```
james-simpson-ST: Just to make sure, if this lives in animated-card.scss will this work applying styles to st-animated-card-iframe which is technically outside the component?

pgs-gmartowski: yes, this is actually not a problem with animated card itself -  there are no styles for iframe which is wrapping this card (which makes card shrinking :/)

pgs-gmartowski: I've set min-width and height i px and 100% of width and height for this element. Should be ok now.

src/components/animated-card/animated-card.scss
```diff
@@ -0,0 +1,8 @@
...
+  overflow: hidden;
+  width: 100%;
```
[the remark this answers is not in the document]
james-simpson-ST: I have removed this from example/style.scss so this is the equivalent of the merchant not doing this. Now it shows the problem

@pgs-gmartowski  can you take a look at it and see why the fix isn't working as intended. (I think it is because animated-card.scss only applies inside the iframe not outside the iframe)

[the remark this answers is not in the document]
pgs-gmartowski: yes, you're right, I made new fix for this, now it attached inline styles to iframe (animated card iframe)

src/core/services/Element.ts
```diff
@@ -68,6 +68,12 @@ export class Element {
     iframe.setAttribute('allowtransparency', 'true');
     iframe.setAttribute('scrolling', 'no');
     iframe.setAttribute('frameborder', '0');
+    if (fieldId === 'st-animated-card-iframe') {
+      iframe.setAttribute(
+        'style',
+        'overflow: hidden; width: 100%; height: 100%; min-height: 300px; min-width: 450px; border: none;'
+      );
+    }
```
pgs-gmartowski: @wirus15, @james-simpson-ST I think this is the only way to style iframe field (which can be changed by merchant styles).

PR #772, opened 2020-02-17
feature/ST-349-add-a-callback-if-validation-fails

pgs-gmartowski:


example/index.html
```diff
@@ -139,12 +159,21 @@ <h1 class="example-form__title">
             var stringified = JSON.stringify(data);
             var testVariable = 'This is what we have got after submit' + stringified;
           };
+          st.successEvent(function() {
+            displayPopup('success-popup', 'This is success message', 'green');
+            console.error('SUCCESS!');
```
wirus15: this can be removed

example/index.html
```diff
@@ -129,6 +128,27 @@ <h1 class="example-form__title">
               }
             });
         }
+
+        function displayPopup(id, text, tp) {
+          var div = document.createElement('div');
```
wirus15: I guess a simple `window.alert()` would be enough 😁

src/components/control-frame/ControlFrame.ts
```diff
@@ -229,10 +232,14 @@ export class ControlFrame extends Frame {
     this._payment
       .processPayment(this._postThreeDRequestTypes, this._card, this._merchantFormData, data)
       .then(() => {
+        this._publishMerchantEvent({
```
wirus15: I think there is no need for another private method here, you can just call 
```typescript
this.messageBus.publish({type: MessageBus.EVENTS_PUBLIC.CALL_MERCHANT_SUCCESS_CALLBACK}, true);
```
It's the same one-line function

src/core/classes/StCodec.class.ts
```diff
@@ -131,6 +131,7 @@ class StCodec {
     StCodec.publishResponse(StCodec._createCommunicationError());
     StCodec._notification.error(Language.translations.COMMUNICATION_ERROR_INVALID_RESPONSE);
     validation.blockForm(FormState.AVAILABLE);
+    this._messageBus.publish({ type: MessageBus.EVENTS_PUBLIC.CALL_MERCHANT_ERROR_CALLBACK }, true);
```
wirus15: That's probably not the best place to put this code, but to do it right it would take a lot of refactoring so let's just leave it there...

src/ST.ts
```diff
@@ -101,6 +101,18 @@ class ST {
     }
   }
 
+  public errorEvent(callback: any) {
```
wirus15: I was about to write that I'd put those callbacks into config to avoid adding more stuff to the ST class, but then I remembered we use json for the config format...

src/core/classes/StCodec.class.ts
pgs-gmartowski: I think if we change our applications architecture, we can create some kind of validation component which returns validation status and message.

example/index.html
pgs-gmartowski: actually ... you're right ;)

example/index.html
pgs-gmartowski: true, done

example/index.html
pgs-gmartowski: done

src/components/control-frame/ControlFrame.ts
pgs-gmartowski: true

src/ST.ts
pgs-gmartowski: :(

example/index.html
```diff
@@ -160,11 +140,11 @@ <h1 class="example-form__title">
             var testVariable = 'This is what we have got after submit' + stringified;
           };
           st.successEvent(function() {
-            displayPopup('success-popup', 'This is success message', 'green');
+            window.alert('Success :)');
```
wirus15: I didn't mean to remove all this code since it was already written :) I leave the choice to you whether you want to use those fancy notifications or `window.alert` :)

pgs-gmartowski: that's ok ;)

pgs-gmartowski: @wirus15 I need to go back to my fancy alert - Marek said window.alert spoils automated tests

example/index.html
```diff
@@ -1,154 +1,185 @@
...
+      }
+    }
+
+    function configurationInit(ST, data) {
+      var stConfig = data;
+      var parsedUrl = new URL(window.location.href);
+      stConfig.jwt = parsedUrl.searchParams.get('jwt') || stConfig.jwt;
+      var st = ST(stConfig);
+
+      st.submitCallback = function someFancyfunction(data) {
```
james-simpson-ST: I thought with the submitCallback we passed this in through the stConfig rather than assigning it to the st object? 

We should standardise the interface here but keep functionality so that it doesn't break any merchant already using submitCallback.

I'm not sure which is the standard approach to being able to provide a library with a callback function (whether it should go in the config or like you've done here where its to pass in a function) but we should decide and go with the most common approach and stick with it going forward (but if necessary making the submitCallback work the original way and any new way we come up with)

pgs-gmartowski: yes, now I see a difference :/ Technically both solutions are ok, but if we want to continue the same approach as for `submitCallback` then we should change current version of merchants callback

wirus15: I'd go with the original approach and keep the same way of assigning callbacks as in the `submitCallback`. Like James said, keep a standarised approach and don't break current functionality. However, we could implement the second approach and let the merchant choose which one suits him better, eg.:
```javascript
var st = ST(stConfig);

st.on('submit', function () {
    ...
});

st.on('error', function (error) { 
    console.error(error);
});

st.off('submit'); // unsubscribe

Comment (wirus15) [reply example/index.html]: Plus we can change it to defaults the new behavior and keep the backwards compatibility by using setters.
```javascript
class ST {
    set submitCallback(callback) {
        this.on('submit', callback);
    }
}

Comment (pgs-gmartowski) [reply example/index.html]: true, done

File: src/ST.ts [hunk 8]
```diff
@@ -44,6 +44,18 @@ class ST {
   private _storage: BrowserLocalStorage;
   private _translation: Translator;
 
+  set submitCallback(callback: any) {
+    this.on('submit', callback);
+  }
+
+  set successEvent(callback: any) {
```
wirus15: Should be `successCallback` and `errorCallback`.

Also instead of `any` we can specify a type: `callback: (...args: any[]) => void` or even better, define what args the callback would receive.

```typescript
set successCallback(callback: (event: SuccessEvent) => void) { ... }

Comment (wirus15) [reply src/ST.ts]: Plus, we need to unsubscribe, when merchant passes `null` as callback
```typescript
if (callback) {
    this.on('success', callback);
} else {
    this.off('success');
}

File: src/ST.ts [hunk 9]
```diff
@@ -52,6 +64,22 @@ class ST {
...
+      submit: MessageBus.EVENTS_PUBLIC.CALL_MERCHANT_SUBMIT_CALLBACK
+    };
+    // @ts-ignore
+    this._messageBus.subscribe(events[event], () => {
+      callback();
+    });
+  }
+
+  public off(event: string) {
+    console.error('test');
```
wirus15: 👍 

or maybe not `error` but `log` 😉

src/core/classes/CardFrames.class.ts
```diff
@@ -356,6 +356,7 @@ export class CardFrames extends RegisterFrames {
   private _subscribeBlockSubmit(): void {
     this.messageBus.subscribe(MessageBus.EVENTS.BLOCK_FORM, (state: FormState) => {
       this._disableSubmitButton(state);
+      console.error(state);
```
wirus15: to removal

src/core/classes/RegisterFrames.class.ts
```diff
@@ -24,14 +24,15 @@ export class RegisterFrames {
...
     styles: IStyles,
     animatedCard: boolean,
-    fieldsToSubmit: string[],
-    submitCallback?: any
+    fieldsToSubmit?: string[]
   ) {
     this.messageBus = new MessageBus();
-    this.fieldsToSubmit =
-      fieldsToSubmit !== null && fieldsToSubmit.length ? fieldsToSubmit : RegisterFrames.COMPLETE_FORM_FIELDS;
+    if (fieldsToSubmit === null) {
```
wirus15:
```typescript
this.fieldsToSubmit = fieldsToSubmit || RegisterFrames.COMPLETE_FORM_FIELDS;
```
shorter 😉

pgs-gmartowski: I always forget that I can use this :p

src/core/classes/CardFrames.class.ts
pgs-gmartowski: done

src/ST.ts
pgs-gmartowski: done

src/ST.ts
pgs-gmartowski: done, however I think that merchant can specify function whatever he like and we should define them like:
`set errorCallback(callback: (...args: any[]) => void) {`

src/ST.ts
```diff
@@ -52,6 +76,22 @@ class ST {
...
+      submit: MessageBus.EVENTS_PUBLIC.CALL_MERCHANT_SUBMIT_CALLBACK
+    };
+    // @ts-ignore
+    this._messageBus.subscribe(events[event], () => {
+      callback();
+    });
+  }
+
+  public off(event: string) {
+    console.log(`Event: ${event} is off`);
```
james-simpson-ST: Looks like ts complains about console.log - do we need this or is it just for debugging?

pgs-gmartowski: this was just a placeholder because `off` function will be  useful only when new MessageBus appears

pgs-gmartowski: I can change this to 
` public off(event: string) {
     return `Event: ${event} is off`;
`

pgs-gmartowski: done

src/ST.ts
pgs-gmartowski: changed once again after discussion

PR #779, opened 2020-02-21
Delete the js-payments-testing now it's not used

james-simpson-ST:


pgs-gmartowski:
@james-simpson-ST I think this is why dependabot requested to add .gitmodules file - https://github.com/SecureTrading/js-payments/issues/780; when we delete js-payments-testing, we'll not need this

PR #789, opened 2020-02-23
Feature/stjs 535 storage

wirus15:
I improved the storage synchronization between merchant's and secure trading's domains. Had a really hard time finding the best sollution here but it looks like there's not much more we can do without rewriting half of the application to use asynchronous storage. 

I also removed all the usages of the native storage and replaced them with out implementation (for both `localStorage` and `sessionStorage`)


james-simpson-ST:
Code looks fine but looks like unit tests are having a hard time with this change, I guess a fair bit of them need updating to mock the new objects or similar changes

PR #790, opened 2020-02-24
Wrong schema configuration

pgs-gmartowski:


src/core/config/schema/ConfigSchema.ts
```diff
@@ -52,15 +52,20 @@ export const ConfigSchema: Joi.JoiObject = Joi.object().keys({
     expirationDate: Joi.string().required(),
     notificationFrame: Joi.string().required(),
     securityCode: Joi.string().required()
-  }),
-  components: ComponentsConfigSchema,
+  }).default({}),
+  components:  Joi.object().keys({
+    defaultPaymentType: Joi.string().allow(''),
+    requestTypes: Joi.array().items(Joi.string().valid('THREEDQUERY', 'AUTH')),
```
james-simpson-ST: Shouldn't this allow all of these options? 'ACCOUNTCHECK', 'AUTH', 'JSINIT', 'RISKDEC', 'SUBSCRIPTION', 'THREEDQUERY'

So it should be 
```javascript
requestTypes: Joi.array().items(
    Joi.string().valid('ACCOUNTCHECK', 'AUTH', 'JSINIT', 'RISKDEC', 'SUBSCRIPTION', 'THREEDQUERY')
  )
```

pgs-gmartowski: true, I forgot that there are couple of places where it has to be added explicitly

pgs-gmartowski: done

src/core/config/schema/ConfigSchema.ts
```diff
@@ -52,15 +52,20 @@ export const ConfigSchema: Joi.JoiObject = Joi.object().keys({
...
     notificationFrame: Joi.string().required(),
     securityCode: Joi.string().required()
-  }),
-  components: ComponentsConfigSchema,
+  }).default({}),
+  components:  Joi.object().keys({
+    defaultPaymentType: Joi.string().allow(''),
+    requestTypes: Joi.array().items(Joi.string().valid('ACCOUNTCHECK', 'AUTH', 'JSINIT', 'RISKDEC', 'SUBSCRIPTION', 'THREEDQUERY')),
+    paymentTypes: Joi.array().items(Joi.string().allow('')),
+    startOnLoad: Joi.boolean().allow('').allow('')
```
wirus15:
```
allow('').allow('')
```

btw, I think `allow()` is not necessary for the boolean. Plus, what exactly `allow('')` does (with empty string)?

wirus15: If it supposed to mean that this option can be empty, I think I saw a `.optional()` in the docs

pgs-gmartowski: oops, yes there is too much allow() ;)

pgs-gmartowski: @wirus15 not exactly, here is discussion in Joi's repo - https://github.com/hapijs/joi/issues/482 - seems like it's really odd solution to make a property `not required`

PR #792, opened 2020-02-24
bugfix/ST-549-updateJWT-doesnt-work-correctly-with-frictionless-3DS-payment

pgs-gmartowski:


src/core/shared/Payment.ts
```diff
@@ -55,8 +55,12 @@ export class Payment {
         payload: { jwt, response }
       } = new StJwt(result.jwt);
       const threeDInitResult = { jwt, response: response[0] };
-      // @ts-ignore
-      this._cardinalCommerceCacheToken = result.response.cachetoken;
+      // We should always use the id from the original cachetoken to link up with the THREEDQUERY
+      // We've already passed this into Cardinal and they have used it for the fingerprint by this point
+      if (this._cardinalCommerceCacheToken === undefined) {
+        // @ts-ignore
+        this._cardinalCommerceCacheToken = result.response.cachetoken;
```
pgs-gmartowski: so it means we shouldn't add a new one when we set new JWT ?

pgs-gmartowski: Look like we have new JWT and old cachetoken; is it ok ?

james-simpson-ST: Yes you are right we have a new JWT and old cachetoken, however the cachetoken is only used for linking things together like a unique reference to link the setup with the THREEDQUERY, so if we just use the same one even with a new JWT that is fine as long as we use the same one between the setup and the THREEDQUERY

james-simpson-ST:
> Greate job @james-simpson-ST 🎉
> 
> Just wondering if the rest of the changes are required or just that one `if` did the trick.
> 
> Plus the unit tests need an update

I believe all the remaining changes are needed, the bin.process was actually an oversight by us and the update.jwt that they've added is something new that didn't exist on the initial integration so we should use it instead of our work around (which was to use the setup again).

tldr; yeah we should keep the rest

PR #797, opened 2020-02-25
Feature/stjs 535 storage

wirus15:
This is a duplicate of https://github.com/SecureTrading/js-payments/pull/789 PR because there was some issues with github and it didn't run the hooks.


src/components/control-frame/ControlFrame.ts
```diff
@@ -67,7 +70,10 @@ export class ControlFrame extends Frame {
   private _threeDQueryResult: any;
   private _validation: Validation;
 
-  constructor() {
+  constructor(
+    private _localStorage: BrowserLocalStorage = Container.get(BrowserLocalStorage),
+    private _sessionStorage: BrowserSessionStorage = Container.get(BrowserSessionStorage),
+  ) {
```
pgs-gmartowski: when I should use `@Inject` and when `Container.get()` ?

src/core/services/storage/AbstractStorage.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  protected abstract getSychronizationEventName(): string;
+
+  private emitStorageEvent(): void {
+    let event: Event;
+
+    try {
+      event = document.createEvent('StorageEvent');
+    } catch (e) {
```
pgs-gmartowski: nth: I think in the never version of ES we can omit parameter in `catch`, not sure if this is in official ES version or proposal

src/components/control-frame/ControlFrame.ts
wirus15: When the class which you're adding the dependencies to is already injectable (exists in the container, and all of its dependencies are also injectable, then you can use the `@Inject` (when injection token is a class name you can omit the annotation). 

Otherwise, when the class using dependencies is created by `new SomeClass(...)` and not coming from the container, then you must use the `Container.get()`. This is just temporary as we should move everything to the container.

src/core/services/storage/AbstractStorage.ts
wirus15: image
🤷‍♂

PR #799, opened 2020-02-27
STJS-536 refactored message bus to use communicator and rxjs

wirus15:


src/core/shared/MessageBus.ts
```diff
@@ -46,125 +52,94 @@ export class MessageBus {
...
-  }
+  public subscribe<T>(
+    next?: (value: IMessageBusEvent<T>) => void,
+    error?: (error: any) => void,
+    complete?: () => void
+  ): Unsubscribable;
 
-  public subscribe(eventType: string, callback: any, subscriber?: string) {
-    subscriber = subscriber || window.name;
+  /** @deprecated use RxJS implementation instead */
```
james-simpson-ST: Do we have a plan to update everywhere so we don't need to keep the deprecated functions?

wirus15: I think I can remove all deprecated methods because there are only a few places where they are used and the change is really simple :)

example/index.html
```diff
@@ -167,13 +167,13 @@ <h1 class="example-form__title">
           st.successCallback = function() {
             displayPopup('success-popup', 'This is success message', 'green');
           };
-
-          st.errorCallback = function() {
+          st.errorCallback = function(err) {
```
pgs-gmartowski: looks like we're not using this parameter anymore

src/ST.ts
```diff
@@ -80,10 +79,10 @@ class ST {
     private _framesHub: FramesHub,
     private _storage: BrowserLocalStorage,
     private _sessionStorage: BrowserSessionStorage,
+    private _messageBus: MessageBus,
```
pgs-gmartowski: shouldn't it be deleted  ?

wirus15: No, why?

pgs-gmartowski: hmm is there any place where we're using it ?

PR #801, opened 2020-02-27
feature/ST-498-add-payment-icon-to-credit-card

pgs-gmartowski:
>


src/components/card-number/CardNumber.ts
```diff
@@ -115,8 +126,81 @@ export class CardNumber extends FormField {
     this.messageBus.publish(messageBusEvent);
   }
 
-  private _getBinLookupDetails = (cardNumber: string) =>
-    iinLookup.lookup(cardNumber).type ? iinLookup.lookup(cardNumber) : undefined;
+  private _getImagePath(type: string): string {
+    switch (type) {
```
wirus15: A map would be easier to maintain and with less code:

```typescript
const iconMap = {
  CardNumber.CARDS_IMAGES_FILES_NAMES.AMEX: CardNumber.CARDS_IMAGES_FILES_NAMES.AMEX,
  ...
};

const icon = iconMap[type];

return icon && `./images/${icon.toLowerCase()}.png` || '';

File: src/components/card-number/CardNumber.ts [hunk 2]
```diff
@@ -115,8 +126,81 @@ export class CardNumber extends FormField {
...
+  private _setIconAttributes(imageURI: string, iconId: string): HTMLElement {
+    let element: HTMLElement = document.createElement('img');
+    element = this._setIconStyles(element);
+    element.setAttribute('src', imageURI);
+    element.setAttribute('id', iconId);
+    return element;
+  }
+
+  private _setIconStyles(element: HTMLElement): HTMLElement {
+    element.style.width = '30px';
```
wirus15: Can we move those to the css file?

src/components/card-number/CardNumber.ts
```diff
@@ -115,8 +126,81 @@ export class CardNumber extends FormField {
...
+    element.style.border = '1px solid #6d6d6d';
+    element.style.borderRadius = '4px';
+    return element;
+  }
+
+  private _setImageIconPlaceholder(type: string, iconId: string): void {
+    const icon: HTMLElement = document.getElementById(iconId);
+    const imageURI: string = this._getImagePath(type);
+
+    if (!icon && !imageURI) {
```
wirus15:
```typescript
if (icon) {
  icon.remove();
}

if (!imageURI) {
  return;
}
```
2 conditions instead of 3 😉

src/core/services/icon/IconMap.ts
```diff
@@ -0,0 +1,21 @@
...
+  maestro: `./images/maestro.png`,
+  mastercard: `./images/mastercard.png`,
+  piba: `./images/piba.png`,
+  visa: `./images/visa.png`
+};
+
+@Service()
+export class IconMap {
+  getUrl(name: string) {
+    // @ts-ignore
```
wirus15: this `ignore` is necessary?

src/core/services/icon/IconFactory.ts
```diff
@@ -0,0 +1,30 @@
+import { Container } from 'typedi';
+import { IconMap } from './IconMap';
+
+export class IconFactory {
+  private _attributes = { alt: '', title: '', ariaLabel: '', id: 'card-icon' };
+  private readonly _icon: HTMLImageElement;
```
wirus15: Don't store the icon here. It's a factory it should create a new instance on each call

src/components/card-number/CardNumber.ts
```diff
@@ -126,82 +127,43 @@ export class CardNumber extends FormField {
...
-    element.style.backgroundColor = '#fff';
-    element.style.border = '1px solid #6d6d6d';
-    element.style.borderRadius = '4px';
-    return element;
-  }
-
-  private _setImageIconPlaceholder(type: string, iconId: string): void {
-    const icon: HTMLElement = document.getElementById(iconId);
-    const imageURI: string = this._getImagePath(type);
+  private _setIconImage(type: string, iconId: string): void {
```
wirus15: It would be much easier if you just simply remove the existing icon element and append a new one.

src/core/services/icon/IconFactory.ts
```diff
@@ -0,0 +1,37 @@
...
+  private _attributes: IIconAttributes = {
+    alt: '',
+    ariaLabel: '',
+    class: 'st-card-icon',
+    id: 'card-icon',
+    src: '',
+    title: ''
+  };
+
+  constructor(private _url: IconMap = Container.get(IconMap)) {
```
wirus15: remove `Container.get(...)`, the service will be injected here automatically

src/core/services/icon/IconFactory.ts
```diff
@@ -0,0 +1,37 @@
+import { Container, Service } from 'typedi';
+import { IconMap } from './IconMap';
+import { IIconAttributes } from '../../models/IIconAttributes';
+
+@Service()
+export class IconFactory {
+  private _attributes: IIconAttributes = {
```
wirus15: why do you need this map? especially when half of the attributes are not even used?

src/core/services/icon/IconFactory.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  private _createIcon(name: string): HTMLImageElement {
+    const icon = document.createElement('img');
+    this._setAttributes(icon, name);
+    return icon;
+  }
+
+  getIcon(name: string): HTMLImageElement {
+    return this._createIcon(name);
```
wirus15: you could contain this whole code inside this one method - there's no need to create private methods here

pgs-gmartowski: I thought it should be separated :/ Never mind, I left only `getIcon() `

src/core/services/icon/IconFactory.ts
pgs-gmartowski: I deleted title attribute, but I think rest of them is obligatory or informative; I would rather leave attributes as follows:
`    alt: '',
    ariaLabel: 'Credit card icon',
    class: 'st-card-icon',
    id: 'card-icon',
    src: '',
`

src/core/services/icon/IconFactory.ts
pgs-gmartowski: ok, done

src/components/card-number/CardNumber.ts
```diff
@@ -22,6 +23,17 @@ export class CardNumber extends FormField {
   private static NO_CVV_CARDS: string[] = ['PIBA'];
   private static STANDARD_CARD_LENGTH: number = 19;
   private static WHITESPACES_DECREASE_NUMBER: number = 2;
+  private static CARDS_IMAGES_FILES_NAMES = {
```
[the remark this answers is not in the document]
pgs-gmartowski: nope, deleted

PR #802, opened 2020-02-28
Hotfix/localstorage in incognito in an iframe

james-simpson-ST:
There seems to be some very odd behaviour if we use chrome in incognito mode with a merchant page inside an iframe where the localStorage isn't set or doesn't allow cross-origin or something.

As a quick work around I have added a check into Translator to check if the local storage exists for merchantTranslations and not to crash if it doesn't. But we should have a proper look into this to see why it isn't allowing localstorage in this case.

Also I have reverted the change to validate the submitFields as a merchant can ask for anything not just those that we default to e.g. they could ask for settleduedate which isn't in our list. So we can only validate that its an array for this one

### DIFF
--- a/src/core/models/constants/ConfigSchema.ts
+++ b/src/core/models/constants/ConfigSchema.ts
@@ -78,21 +78,7 @@ export const ConfigSchema: Joi.JoiObject = Joi.object().keys({
   ),
   styles: Joi.object(),
   submitCallback: Joi.any(),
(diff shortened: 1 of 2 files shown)


wirus15:
I did something similar here, but it's not merged yet.
https://github.com/SecureTrading/js-payments/pull/797/files#diff-62e73f7b2d3cf26296b6c4837d5d1da3R49

PR #811, opened 2020-03-03
[WIP] clean-config-file

pgs-gmartowski:


pgs-gmartowski:
@james-simpson-ST, @wirus15 I've improved config file - right now we have all of supported properties which can be specified in config.json. Couple of them seems to be duplicated :/

James, could you take a look at config.json file and let me know if we can delete some of properties (eg. requestTypes which has triple occurence)

src/core/classes/CardFrames.class.ts
```diff
@@ -228,10 +220,10 @@ export class CardFrames extends RegisterFrames {
   private _initAnimatedCardFrame(): void {
     this._animatedCard = new Element();
     const animatedCardConfig = { ...this.params };
-    if (this._paymentTypes !== undefined) {
+    if (typeof this._paymentTypes !== 'undefined' && this._paymentTypes) {
```
wirus15: `if (this._paymentTypes) { ... }` would be enough :)

src/core/config/ConfigResolver.ts
```diff
@@ -52,49 +52,57 @@ export class ConfigResolver {
...
       visaCheckout: this._setApmConfig(config.visaCheckout, config.components),
       ...this._setFieldsToSubmit(config),
       ...this._setPropertiesToSubmit(config)
     };
   }
 
+  private _isTruthy = (value: any) => {
+    const valueType = typeof value;
+    if (valueType === 'object') {
+      return typeof value !== 'undefined' && value.length > 0;
```
wirus15: you don't need to check if `typeof` is `undefined` because you've already checked if it's `object`

src/core/config/ConfigResolver.ts
```diff
@@ -52,49 +52,57 @@ export class ConfigResolver {
...
+    }
+    return typeof value !== 'undefined' && value;
```
wirus15: `return Boolean(value);`

pgs-gmartowski:
I'm closing this PR and continue topic in https://github.com/SecureTrading/js-payments/pull/863

PR #812, opened 2020-03-03
[WIP] depcheck

pgs-gmartowski:


pgs-gmartowski:
@wirus15, @msakowicz - it might fail from time to time - It'll throw an error if some dependency seems to be unused, or at least will be marked as used

pgs-gmartowski:
Needs to be closed because of vulnerabilities discovered in npm-check package

PR #816, opened 2020-03-04
feature/ST-545-allow-merchant-to-not-include-notification-frame

pgs-gmartowski:


src/core/shared/Notification.ts
```diff
@@ -2,34 +2,46 @@ import { NotificationType } from '../models/constants/NotificationType';
 import { IMessageBusEvent } from '../models/IMessageBusEvent';
 import { INotificationEvent } from '../models/INotificationEvent';
 import { MessageBus } from './MessageBus';
+import { Service } from 'typedi';
```
james-simpson-ST: Nice another re-used dependency done 😄

james-simpson-ST:
We need to figure out why behave tests are failing but changes all look good and shouldn't have caused failures so not sure why

src/core/config/ConfigResolver.ts
```diff
@@ -52,51 +52,60 @@ export class ConfigResolver {
...
+      livestatus: this._isTruthy(config.livestatus) ? config.livestatus : 0,
+      notifications: this._isTruthy(config.notifications) ? config.notifications : false,
+      origin: this._isTruthy(config.origin) ? config.origin : window.location.origin,
+      panIcon: this._isTruthy(config.panIcon) ? config.panIcon : false,
+      placeholders: this._isTruthy(config.placeholders) ? config.placeholders : { pan: '', expirydate: '', securitycode: '' },
+      styles: this._isTruthy(config.styles) ? config.styles : {},
+      submitCallback: this._isTruthy(config.submitCallback) ? config.submitCallback : null,
+      submitFields: this._isTruthy(config.submitFields) ? config.submitFields : [],
+      submitOnError: this._isTruthy(config.submitOnError) ? config.submitOnError : false,
+      submitOnSuccess: this._isTruthy(config.submitOnSuccess) ? config.submitOnSuccess : false,
```
james-simpson-ST: I think it is this which is different so should default to true

james-simpson-ST: I've changed this back so let's see how tests handle now 😄

james-simpson-ST: Okay tests didn't pass after ensuring the defaults are backwards compatible, @pgs-gmartowski and @msakowicz can you continue to debug

pgs-gmartowski: @james-simpson-ST yes, I think sth is wrong with default config - I have the same issues in clean-config-file branch. Looks like default config is not set in ConfigResolver. I must speak with Maciej `bout this :/

src/components/control-frame/ControlFrame.ts
```diff
@@ -406,7 +406,7 @@ export class ControlFrame extends Frame {
   private _setInstances(): void {
     this._payment = new Payment(this.params.jwt, this.params.gatewayUrl, this.params.origin);
     this._validation = new Validation();
-    this._notification = new Notification();
+    this._notification = Container.get(Notification);
```
wirus15: I think this can be already injected into constructor

src/core/config/ConfigResolver.ts
```diff
@@ -52,51 +52,62 @@ export class ConfigResolver {
...
       visaCheckout: this._setApmConfig(config.visaCheckout, config.components),
       ...this._setFieldsToSubmit(config),
       ...this._setPropertiesToSubmit(config)
     };
   }
 
+  private _isTruthy = (value: any) => {
+    const valueType = typeof value;
+    if (valueType === 'object') {
+      return Object.keys(value).length;
```
wirus15: this function should return `boolean`, so `return Boolean(Object.keys(value).length);`

src/core/config/ConfigResolver.ts
```diff
@@ -52,51 +52,62 @@ export class ConfigResolver {
...
+      translations: this._isTruthy(config.translations) ? config.translations : {},
       visaCheckout: this._setApmConfig(config.visaCheckout, config.components),
       ...this._setFieldsToSubmit(config),
       ...this._setPropertiesToSubmit(config)
     };
   }
 
+  private _isTruthy = (value: any) => {
+    const valueType = typeof value;
+    if (valueType === 'object') {
```
wirus15: you should do the same check for arrays

src/core/config/ConfigResolver.ts
```diff
@@ -52,51 +52,62 @@ export class ConfigResolver {
...
+      return Object.keys(value).length;
+    }
+    return valueType !== 'undefined' && value;
```
wirus15: here you can simply return `Boolean(value)`

src/core/config/ConfigResolver.ts
```diff
@@ -52,51 +52,62 @@ export class ConfigResolver {
...
+
   private _componentIds(config: IComponentsIds): IComponentsIds {
-    if (!config) {
+    if (!this._isTruthy(config)) {
       return { ...this.DEFAULT_COMPONENTS_IDS };
     }
-
-    const optionalIds = config.animatedCard !== undefined ? { animatedCard: config.animatedCard } : {};
+    const optionalIds = {
+      animatedCard: this._isTruthy(config.animatedCard) ? config.animatedCard : this.DEFAULT_COMPONENTS_IDS.animatedCard
```
wirus15: I'm thinking about a helper method here: `animatedCard: this.getValueOrDefault(config.animatedCard, this.DEFAULT_COMPONENTS_IDS.animatedCard)`

src/core/config/ConfigResolver.ts
```diff
@@ -52,51 +52,62 @@ export class ConfigResolver {
...
-      submitFields: config.submitFields !== undefined ? config.submitFields : [],
-      submitOnError: config.submitOnError !== undefined ? config.submitOnError : false,
+      datacenterurl: this._isTruthy(config.datacenterurl) ? config.datacenterurl : environment.GATEWAY_URL,
+      deferInit: this._isTruthy(config.deferInit) ? config.deferInit : false,
+      formId: this._isTruthy(config.formId) ? config.formId : Selectors.MERCHANT_FORM_SELECTOR,
+      init: this._isTruthy(config.init) ? config.init : { cachetoken: '', threedinit: '' },
+      jwt: this._isTruthy(config.jwt) ? config.jwt : '',
+      livestatus: this._isTruthy(config.livestatus) ? config.livestatus : 0,
+      // Can't use isTruthy for notifications because we default to true so undefined wants to act as if it's truthy
+      notifications: config.notifications !== undefined ? config.notifications : true,
```
wirus15: you can change this property to `disableNotification` which is `false` by default

src/core/shared/Notification.ts
```diff
@@ -2,34 +2,46 @@ import { NotificationType } from '../models/constants/NotificationType';
...
   private _messageBusEvent: IMessageBusEvent;
   private _notificationEvent: INotificationEvent;
+  private readonly _display: boolean;
+  private readonly _displayOnError: boolean;
+  private readonly _displayOnSuccess: boolean;
 
-  constructor() {
+  constructor(private _configProvider: ConfigProvider) {
     this._messageBus = Container.get(MessageBus);
+    this._display = this._configProvider.getConfig().notifications;
```
wirus15: You say that there is a problem that the configuration is empty. This might be because in the moment of creation of this class the config is still not set. Try changing this properties to be lazy loaded from the provider instead of getting them eagerly in the constructor. You can do this by calling the `configProvider` for config inside `error/info/success/warning` methods or by changing the properties to getters.

test/components/card-number/CardNumber.spec.ts
```diff
@@ -12,6 +12,7 @@ import { IconFactory } from '../../../src/core/services/icon/IconFactory';
 
 jest.mock('./../../../src/core/shared/MessageBus');
 jest.mock('./../../../src/core/shared/Validation');
+jest.mock('./../../../src/core/shared/Notification');
```
wirus15: why do you put this mock here if nothing else changed in this spec file?

src/core/config/ConfigResolver.ts
pgs-gmartowski: hmm, but `typeof` on array gives us 'object'; that's why I check `value.length` first, then `Object.keys(value).length`

src/core/config/ConfigResolver.ts
pgs-gmartowski: true, fixed

src/components/control-frame/ControlFrame.ts
pgs-gmartowski: true, done

src/core/config/ConfigResolver.ts
pgs-gmartowski: xD true :)

pgs-gmartowski: done

src/core/config/ConfigResolver.ts
pgs-gmartowski: yes, now it's much more neat :)

test/components/card-number/CardNumber.spec.ts
pgs-gmartowski: don't know xD deleted

src/core/shared/Notification.ts
pgs-gmartowski: true, changed

src/core/config/ConfigResolver.ts
pgs-gmartowski: I think this has to stay the same because it's part of API

src/core/config/ConfigResolver.ts
pgs-gmartowski: anyway, I changed this function like you mentioned and now should do what we exactly wanted

wirus15: > hmm, but `typeof` on array gives us 'object'; that's why I check `value.length` first, then `Object.keys(value).length`

https://developer.mozilla.org/pl/docs/Web/JavaScript/Referencje/Obiekty/Array/isArray 😉

src/core/config/ConfigResolver.ts
wirus15: Yes, but it's a new property so you can make it whatever you want :)

src/core/config/ConfigResolver.ts
```diff
@@ -52,91 +52,75 @@ export class ConfigResolver {
...
     const valueType = typeof value;
-    if (valueType === 'object') {
-      if (value.length) {
-        return true;
-      } else {
-        return Boolean(Object.keys(value).length);
-      }
+    if (valueType === 'object' && (Boolean(value.length) || Boolean(Object.keys(value).length))) {
+      return value;
+    } else if (Boolean(value)) {
```
wirus15: skip the `else`'s
```typescript
if (valueType === 'object' && (Boolean(value.length) || Boolean(Object.keys(value).length))) {
  return value;
}

if (Boolean(value)) {
  return value;
}

return defaultValue;

File: src/core/shared/Notification.ts [hunk 12]
```diff
@@ -7,28 +7,56 @@ import { ConfigProvider } from '../config/ConfigProvider';
 
 @Service()
 export class Notification {
+
+  get display(): boolean {
```
wirus15: you got this wrong, by adding those getters and setters you created public properties and thats not what we want. You should have created private getters that returns values from the config.

```typescript
private get _displayNotifications(): boolean {
  return this._configProvider.getConfig().notifications;
}

private get _submitOnError(): boolean {
  return this._configProvider.getConfig().submitOnError;
}

public error(message: string): void {
    const displyOnError = this._display && !this._submitOnError;
    this._setNotification(NotificationType.Error, message, displayOnError);
}

File: src/core/shared/Notification.ts [hunk 13]
```diff
@@ -7,28 +7,56 @@ import { ConfigProvider } from '../config/ConfigProvider';
...
 
   public info(message: string) {
     this._setNotification(NotificationType.Info, message);
   }
 
   public success(message: string) {
-    this._setNotification(NotificationType.Success, message, this._displayOnSuccess);
+    this.display = this._configProvider.getConfig().notifications;
+    this.displayOnSuccess = this.display && !this._configProvider.getConfig().submitOnSuccess;
+    console.error(this.displayOnSuccess);
```
wirus15: remove those

src/core/config/ConfigResolver.ts
pgs-gmartowski: true, done

src/core/config/ConfigResolver.ts
pgs-gmartowski: that make sense 😃

src/core/config/ConfigResolver.ts
pgs-gmartowski: it has changed a little bit since that implementation, please look at this now

src/core/shared/Notification.ts
pgs-gmartowski: true, I've made changes

src/core/config/ConfigResolver.ts
```diff
@@ -86,13 +86,13 @@ export class ConfigResolver {
 
   private _getValueOrDefault(value: any, defaultValue: any) {
     const valueType = typeof value;
-    if (valueType === 'object' && (Boolean(value.length) || Boolean(Object.keys(value).length))) {
+    if (valueType === 'object' && (Array.isArray(value) || Boolean(Object.keys(value).length))) {
```
wirus15: right now you can simplify this function even further:
```typescript
private _getValueOrDefault(value: any, defaultValue: any) {
  if (typeof(value) !== 'undefined') {
    return value;
  }

  return defaultValue;
}
```
I'm also wondering if generic typing would work here:
```typescript
private _getValueOrDefault<T>(value: T | undefined, defaultValue: T): T {
  if (typeof(value) !== 'undefined') {
    return value;
  }

  return defaultValue;
}

src/core/shared/Notification.ts
pgs-gmartowski: done

src/core/config/ConfigResolver.ts
pgs-gmartowski: done

pgs-gmartowski: looks like it's workin' even in this second approach

PR #818, opened 2020-03-05
[WIP] hotfix/ST-554-potential-xss-if-styling-passed-in

pgs-gmartowski:


src/core/shared/DomMethods.ts
```diff
@@ -81,16 +81,18 @@ export class DomMethods {
...
       return;
     }
-    const style = document.createElement(DomMethods.STYLE_MARKUP);
     style.setAttribute('id', 'insertedStyles');
     style.setAttribute('type', 'text/css');
-    style.innerHTML = contents;
-    document.head.appendChild(style);
+    const index = document.styleSheets.length - 1;
+    contents.forEach((item: string) => {
+      (document.styleSheets[index] as CSSStyleSheet).insertRule(item, 0);
```
pgs-gmartowski: @james-simpson-ST I've checked this and works like a charm ;)

src/core/shared/DomMethods.ts
```diff
@@ -81,16 +81,18 @@ export class DomMethods {
...
+    const style = document.createElement(DomMethods.STYLE_MARKUP);
     if (head) {
       return;
     }
-    const style = document.createElement(DomMethods.STYLE_MARKUP);
     style.setAttribute('id', 'insertedStyles');
     style.setAttribute('type', 'text/css');
-    style.innerHTML = contents;
-    document.head.appendChild(style);
+    const index = document.styleSheets.length - 1;
```
james-simpson-ST: We don't seem to be using the index? Do we need to or are we okay specifying 0 always as they are independent rules anyway that won't clash

pgs-gmartowski: actually you're right; we don't have anything more than one stylesheet for each component. 
Fixed.

src/core/shared/DomMethods.ts
wirus15:
```typescript
const styleSheet: CSSStyleSheet = document.styleSheets[document.styleSheets.length - 1];
contents.forEach((item: string) => styleSheet.insertRule(item, 0));
```
😉

src/core/shared/DomMethods.ts
```diff
@@ -81,16 +81,17 @@ export class DomMethods {
...
     if (head) {
       return;
     }
-    const style = document.createElement(DomMethods.STYLE_MARKUP);
     style.setAttribute('id', 'insertedStyles');
     style.setAttribute('type', 'text/css');
-    style.innerHTML = contents;
-    document.head.appendChild(style);
+    contents.forEach((item: string) => {
+      (document.styleSheets[0] as CSSStyleSheet).insertRule(item, 0);
```
james-simpson-ST: I just realised what the index was for >.< it now gets added to the main.css instead of the newly created style element

I think this might be better:

```javascript
const head = document.getElementById('insertedStyles');
    const style = document.createElement(DomMethods.STYLE_MARKUP) as HTMLStyleElement;
    if (head) {
      return;
    }
    style.setAttribute('id', 'insertedStyles');
    style.setAttribute('type', 'text/css');
    contents.forEach((item: string) => {
      (style.sheet as CSSStyleSheet).insertRule(item, 0);
    });
```

james-simpson-ST: Sorry I missed the `document.head.appendChild(style)` that needs adding back in (updated above)

pgs-gmartowski: good point ! Looks much better now :)

pgs-gmartowski: @wirus15, @james-simpson-ST I think both your comments have the same notice. Please take a look at my changes.

src/core/shared/DomMethods.ts
pgs-gmartowski: fixed

PR #820, opened 2020-03-06
STJS-551 removes usage of window object directly when identifying frames

wirus15:
In order to make the app work inside in iframe it was necessary to change the way the frames are identified, like:
- using `window.top`
- checking if `window.name` is empty


src/core/services/message-bus/FrameIdentifier.spec.ts
```diff
@@ -0,0 +1,61 @@
+import { FrameIdentifier } from './FrameIdentifier';
```
james-simpson-ST: This is in the wrong place, shouldn't it be in /tests/core/services/message-bus?

tslint.json
```diff
@@ -12,7 +12,7 @@
     "no-conditional-assignment": true,
     "no-consecutive-blank-lines": false,
     "no-console": {
-      "severity": "warning",
+      "severity": "off",
```
james-simpson-ST: I don't think we want to turn this off, do we?

src/core/services/message-bus/FrameIdentifier.spec.ts
wirus15: @james-simpson-ST I proposed this to @pgs-gmartowski and we agreed that this would be a better approach. 

By keeping the tests in the same folder as the class, we can already see if the class has any unit tests. Plus they are easier to find and when you move the file, you also move the tests for it, so the directory tree in `src` and `test` don't diverge.

The same approach is proposed and used by popular frameworks like Angular.

PR #836, opened 2020-03-11
[WIP] Feature/stjs 569 folder structure

wirus15:
Main changes:
- split application into 3 main folders:
  - `frontend` - merchant's side - ST class and others required to create the iframes
  - `backend` - core (control frame) + other components
  - `shared` - everything used on both sides
- moved unit tests files to the same folder as classes they test

@pgs-gmartowski @james-simpson-ST 
Tell me what you think :) I'm also open for any proposals about new names for main folders (frontend/backend) 😁


wirus15:
image

james-simpson-ST:
I don't have an issue with the tests being in line as long as the compiled code does not contain any tests or test data and that the regexs for finding the tests cover looking in src

I'm not a fan of frontend/backend names because backend makes me think of a server which in this case it will still be executed on the client. However, I'm not sure what a good naming convention would be. Maybe client and application but I'm not 100% sure about that either.

pgs-gmartowski:
> I don't have an issue with the tests being in line as long as the compiled code does not contain any tests or test data and that the regexs for finding the tests cover looking in src
> 
> I'm not a fan of frontend/backend names because backend makes me think of a server which in this case it will still be executed on the client. However, I'm not sure what a good naming convention would be. Maybe client and application but I'm not 100% sure about that either.

I had the same thoughts; I think there is no good naming convention in this case; maybe sth like: ST / Merchant ? Anyway I think this will be a temporary solution utill we split js-payments into client and ST part.

wirus15:
@james-simpson-ST the test finder is looking for files named `*.spec.ts` and the compiler only includes the source files that are imported in other files, so unless we import a test file into a source file, we're good 😁 

I like the names `client` and `application`, that was one of my ideas too 👍

package.json
```diff
@@ -24,8 +24,8 @@
     "snyk": "snyk test --org=secure-trading",
     "testdocker": "docker pull \"securetrading1/js-payments-testing:$DOCKER_BRANCH\" && docker run -d -p 8443:8443 -p 8760:8760 -it \"securetrading1/js-payments-testing:$DOCKER_BRANCH\"",
     "behave": "npm run testdocker && cd js-payments-testing &&",
-    "export:translations": "cross-env-shell i18next-conv -l '$npm_package_config_locale' -s ./src/core/translations/$npm_package_config_locale.json -t ./translations/$npm_package_config_locale.po",
-    "import:translations": "cross-env-shell i18next-conv -l '$npm_package_config_locale' -s ./translations/$npm_package_config_locale.po -t ./src/core/translations/$npm_package_config_locale.json",
+    "export:translations": "cross-env-shell i18next-conv -l '$npm_package_config_locale' -s src/backend/core/translations/$npm_package_config_locale.json -t src/translations/$npm_package_config_locale.po",
```
james-simpson-ST: Don't these also need updating

webpack.common.js
```diff
@@ -8,9 +8,9 @@ const CopyPlugin = require('copy-webpack-plugin');
 
 module.exports = {
   entry: {
-    main: ['./polyfills', './src/components/index.ts'],
-    'control-frame': ['./polyfills', './src/components/control-frame/control-frame.ts'],
-    st: ['./polyfills', './src/ST.ts'],
+    st: ['./polyfills', './src/frontend/ST.ts'],
+    main: ['./polyfills', './src/backend/components/index.ts'],
```
james-simpson-ST: Don't these also need updating

james-simpson-ST:
Good luck with the merge conflicts ;)

PR #837, opened 2020-03-11
[WIP] feature/ST-513-migration-from-tslint-to-eslint

pgs-gmartowski:


pgs-gmartowski:
@wirus15 I think there might be a problem with decorators when using eslint; do you know any workaround / plugin for this ?

tsconfig.json
```diff
@@ -1,28 +0,0 @@
-{
```
wirus15: I think this file is necessary

webpack.common.js
```diff
@@ -137,37 +138,23 @@ module.exports = {
         use: ['file-loader']
       },
       {
-        test: /\.tsx?|js$/,
```
wirus15: It seems that you removed the TS compilation completely? 😁

tsconfig.json
pgs-gmartowski: true :/

webpack.common.js
pgs-gmartowski: yes :/ I need to set it all once again :/

PR #864, opened 2020-03-25
Feature/stjs 576 update jwt problems

wirus15:
- fixed duplicated cardinal requests
- fixed disappearing visa button

@james-simpson-ST I started this branch off the develop but I'm wondering if it could be changed to the release? @msakowicz says those issues also exists on release branch


james-simpson-ST:
> * fixed duplicated cardinal requests
> * fixed disappearing visa button
> 
> @james-simpson-ST I started this branch off the develop but I'm wondering if it could be changed to the release? @msakowicz says those issues also exists on release branch

Changes look quite simple so probably worth having on release branch if we can and we're confident there's no potential issues with it which I feel they're probably okay seeing as behave tests are passing with these changes afaik

PR #867, opened 2020-03-27
STJS-589 fixed submit callback set with config, added other callbacks to config

wirus15:


src/core/config/ConfigResolver.ts
```diff
@@ -78,6 +78,8 @@ export class ConfigResolver {
       requestTypes: this._getValueOrDefault(config.requestTypes, []),
       styles: this._getValueOrDefault(config.styles,{}),
       submitCallback: this._getValueOrDefault(config.submitCallback, null),
+      successCallback: this._getValueOrDefault(config.successCallback, null),
+      errorCallback: this._getValueOrDefault(config.errorCallback, null),
```
pgs-gmartowski: we need to be careful with ConfigResolver - I've found that in full config implementation it's not setting default requestTypes - please test this in various of configurations.

james-simpson-ST:
lgtm; on develop branch once this is merged back I think we should change example/index.html to pass functions into the config instead of setting them on ST instance. I have added task to the top of the backlog for this for next sprint

PR #883, opened 2020-03-30
release/2.0.9

james-simpson-ST:


src/core/classes/StCodec.class.ts
```diff
@@ -180,7 +175,9 @@ class StCodec {
         StCodec.publishResponse(responseContent, jwtResponse);
         StCodec.getNotification().error(responseContent.errormessage);
         StCodec.getMessageBus().publish({ type: MessageBus.EVENTS_PUBLIC.CALL_MERCHANT_ERROR_CALLBACK }, true);
-        throw new Error(responseContent.errormessage);
+        if (!(responseContent.walletsource && responseContent.walletsource === 'APPLEPAY')) {
```
james-simpson-ST, line 178: Doesn't this now mean we're calling publishingResponse twice in the case of APPLEPAY?

I thought the code was meant to catch this error in the StTransport.class.ts sendRequestInternal in which case it wouldn't be a problem or is this catch not working as expected?

pgs-gmartowski: Good point ! I didn't notice StCodec.publishResponse before. 
Actually when we throw an error here, somehow response is not passed further to next functions; that's why at the end in ApplePays  `_onPaymentAuthorized`  we cannot call completePayment function which closes AP popup.

src/components/security-code/SecurityCode.ts
```diff
@@ -134,6 +134,7 @@ export class SecurityCode extends FormField {
     });
 
     this.messageBus.subscribe(MessageBus.EVENTS.IS_CARD_WITHOUT_CVV, (state: FormState) => {
+      console.error(state);
```
james-simpson-ST, line 137: Should we remove this, is it just for debugging?

src/core/classes/StCodec.class.ts
pgs-gmartowski: I've changed implementation a little bit; looks like now it's working fine.

src/components/security-code/SecurityCode.ts
pgs-gmartowski: oops, sorry :) yes, I forgot about this

pgs-gmartowski: done

PR #884, opened 2020-03-30
bugfix/build-issues-on-develop

pgs-gmartowski:


src/client/config/ConfigResolver.ts
```diff
@@ -60,7 +54,7 @@ export class ConfigResolver {
       cachetoken: this._getValueOrDefault(config.cachetoken, ''),
       componentIds: this._componentIds(config.componentIds),
       components: this._setComponentsProperties(config),
-      datacenterurl: this._getValueOrDefault(config.datacenterurl, environment.GATEWAY_URL),
+      datacenterurl: this._getValueOrDefault(config.datacenterurl, environment.GATEWAY_URL) || environment.GATEWAY_URL,
```
wirus15: the `|| environment.GATEWAY_URL` is not necessary after fixing `_getValueOrDefault()`

src/client/config/ConfigResolver.ts
```diff
@@ -73,45 +67,59 @@ export class ConfigResolver {
...
       translations: this._getValueOrDefault(config.translations, {}),
-      visaCheckout: this._setApmConfig(config.visaCheckout, config.components)
+      visaCheckout: this._setApmConfig(config.visaCheckout, config.components),
     };
   }
 
   private _getValueOrDefault<T>(value: T | undefined, defaultValue: T): T {
-    if (typeof (value) !== 'undefined') {
+    const type = typeof value;
+    if (type !== 'undefined') {
```
wirus15: you can change this to early exit:
```typescript
if (type === undefined) {
  return defaultValue;
}
```
you can also use switch here:
```typescript
switch(type) {
  case 'undefined':
    ...
  case 'string':
    ...
  case 'object':
    ...
  default:
}

PR #885, opened 2020-03-30
STJS-595 - Fix ApplePay publish event

james-simpson-ST:
* Fix the publish to always publish to parent as it was not getting the parentOrigin set correctly, but we don't need it anymore
* Also removed the unnecessary argument
* Reduced timeout it takes to detect applypay is complete as the local storage shouldn't take longer than 500ms. 
* Fixed typo of event name which triggers the success callback rather than the error callback for ApplePay

### DIFF
--- a/src/components/control-frame/ControlFrame.ts
+++ b/src/components/control-frame/ControlFrame.ts
@@ -404,7 +404,7 @@ export class ControlFrame extends Frame {
   }
 
   private _setInstances(): void {
-    this._payment = new Payment(this.params.jwt, this.params.gatewayUrl, this.params.origin);
+    this._payment = new Payment(this.params.jwt, this.params.gatewayUrl);
     this._validation = new Validation();
   }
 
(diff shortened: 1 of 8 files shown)


src/core/integrations/ApplePay.ts
```diff
@@ -448,7 +442,7 @@ export class ApplePay {
 
   private _displayNotification(errorcode: string) {
     if (errorcode === '0') {
-      this._messageBus.publish({ type: MessageBus.EVENTS_PUBLIC.CALL_MERCHANT_ERROR_CALLBACK }, true);
+      this._messageBus.publish({ type: MessageBus.EVENTS_PUBLIC.CALL_MERCHANT_SUCCESS_CALLBACK }, true);
```
pgs-gmartowski: good point ! 
I'm not sure but same bug can be in VisaCheckout :/

src/components/control-frame/ControlFrame.ts
```diff
@@ -404,7 +404,7 @@ export class ControlFrame extends Frame {
   }
 
   private _setInstances(): void {
-    this._payment = new Payment(this.params.jwt, this.params.gatewayUrl, this.params.origin);
+    this._payment = new Payment(this.params.jwt, this.params.gatewayUrl);
```
pgs-gmartowski: 👍 that make sense

src/core/integrations/ApplePay.ts
james-simpson-ST: I checked and it didn't have the same typo 👍 so Visa is okay

PR #887, opened 2020-04-01
feature/ST-567-ensure-that-full-tests-are-stable

msakowicz:


james-simpson-ST:
I wonder should we actually also do this on the master? Could you create a separate PR with the same changes but from the master instead of from develop? I don't see why we don't want to do it there too

msakowicz:
sure, I can do it. 
FYI, probably, some of jobs will still failing on master branch. On develop we have few additional fixes (e.g. solved bug with recognize PIBA card icon on IE, etc) But anyway, dividing jobs should solve problem with travis time limitation

PR #889, opened 2020-04-01
STJS-575 cybertonica tid passed in requests

wirus15:


src/application/components/control-frame/ControlFrame.ts
```diff
@@ -416,4 +419,12 @@ export class ControlFrame extends Frame {
   private _updateMerchantFields(data: IMerchantData): void {
     this._merchantFormData = data;
   }
+
+  private _initCybertonica(): void {
+    const config = this._configProvider.getConfig();
```
pgs-gmartowski, line 424: why not: 
`const { cybertonicaApiKey }= this._configProvider.getConfig();`
?

src/application/core/integrations/Cybertonica.ts
```diff
@@ -0,0 +1,39 @@
...
+import { IAFCybertonica } from '../models/cybertonica/IAFCybertonica';
+import { environment } from '../../../environments/environment';
+import { BrowserLocalStorage } from '../../../shared/services/storage/BrowserLocalStorage';
+
+declare const AFCYBERTONICA: IAFCybertonica;
+
+@Service()
+export class Cybertonica {
+  private static readonly SDK_ADDRESS = environment.CYBERTONICA.CYBERTONICA_LIVE_URL;
+  private static ERROR_KEY: string = 'An error occurred';
```
pgs-gmartowski, line 13: you can get this message from translations

src/shared/model/config/IConfig.ts
```diff
@@ -37,4 +37,5 @@ export interface IConfig {
   threedinit?: string;
   translations?: {};
   visaCheckout?: IWalletConfig;
+  cybertonicaApiKey: string;
```
pgs-gmartowski, line 40: should `cybertonicaApiKey` be required ?

example/json/config.json
```diff
@@ -108,5 +108,6 @@
     "settings": {
       "displayName": "My Test Site"
     }
-  }
+  },
+  "cybertonicaApiKey": "test"
```
james-simpson-ST, line 112: With this in the default config would we be constantly connecting to Cybertonica's live system? In which case our automated tests might hit their system a lot. I wonder if we should have a mock like ApplePay/VisaCheckout?

src/environments/environment.prod.ts
```diff
@@ -23,6 +23,9 @@ export const environment = {
     TEST_BUTTON_URL: 'https://sandbox.secure.checkout.visa.com/wallet-services-web/xo/button.png',
     TEST_SDK: 'https://sandbox-assets.secure.checkout.visa.com/checkout-widget/resources/js/integration/v1/sdk.js'
   },
+  CYBERTONICA: {
+    CYBERTONICA_LIVE_URL: 'https://p.cybertonica.com/js/beacon.min.js'
```
james-simpson-ST, line 27: I've just checked the latest document and this URL seams to have changed for a newer version. I've attached the latest Cybertonica document to the Jira story and it is described in Section 2 "Client-side module"

It's worth double checking if anything else has changed in this version too for the client side module

src/application/core/models/cybertonica/IAFCybertonica.ts
```diff
@@ -0,0 +1,16 @@
+export interface IAFCybertonica {
```
james-simpson-ST, line 1: If Cybertonica start returning a new field without telling us will this still work or will it complain? I guess because it's typescript and only during compilation it won't cause issues?

src/application/core/shared/Payment.ts
```diff
@@ -41,13 +42,15 @@ export class Payment {
...
-    this._processPaymentRequestBody = Object.assign(
-      { requesttypedescriptions: requestTypes },
-      additionalData,
-      merchantData,
-      payment
-    );
-    return await this._stTransport.sendRequest(this._processPaymentRequestBody).then(({ response }: any) => response);
+    const processPaymentRequestBody = {
+      requesttypedescriptions: requestTypes,
+      fraudcontroltransactionid: await this._cybertonica.getTransactionId(),
```
james-simpson-ST, line 47: If cybertonica isn't given a cybertonicaApiKey will this return undefined? In which case I think we should only send the key if it exists as I believe the backend will complain if the key exists with a null in the JSON request in an upcoming change to the backend

src/application/core/shared/Payment.ts
```diff
@@ -8,19 +8,20 @@ import { StJwt } from './StJwt';
...
   private _stTransport: StTransport;
   private _threeDInitRequestBody: IStRequest;
-  private _threeDQueryRequestBody: IStRequest;
   private _validation: Validation;
+  private _cybertonica: Cybertonica;
   private readonly _walletVerifyRequest: IStRequest;
 
   constructor(jwt: string, gatewayUrl: string, parentOrigin?: string) {
     this._notification = Container.get(NotificationService);
+    this._cybertonica = Container.get(Cybertonica);
```
james-simpson-ST, line 24: I really like the way this DI is simplifying things :)

example/json/config.json
wirus15: Aren't the smoke tests using their own config for each test case? If the `cybertonicaApiKey` is not set then the service is not initalizing and not performing any requests.

src/application/core/models/cybertonica/IAFCybertonica.ts
wirus15: This won't impact anything as the interface is only being used for typecheck during compilation. It also informs the developer what methods are available in their service. Unless they remove any methods we use, we're fine :)

src/application/core/shared/Payment.ts
wirus15: I was about to ask you about that. Do we know if the backend will behave like you said? When we pass any additional params that are empty the request will be rejected?

example/json/config.json
james-simpson-ST: We have the task STJS-405 which will add a test case for cybertonica so we probably want it for this task? Should it be done there or done now?

src/application/core/shared/Payment.ts
james-simpson-ST: Actually I have just tested this and null is actually fine and ignored by the backend

PR #895, opened 2020-04-03
feature/STJS-602-config-resolver-class-default-config-check

pgs-gmartowski:


src/application/core/models/IVisaSettings.ts
```diff
@@ -1,3 +1,26 @@
...
-  [key: string]: string;
+  locale?: string;
+  countryCode?: string;
+  displayName?: string;
+  websiteUrl?: string;
+  customerSupportUrl?: string;
+  enableUserDataPrefill?: boolean;
+  shipping?: {
+    acceptedRegions?: string[];
+    collectShipping?: 'true' | 'false';
```
wirus15, line 10: those are string values?

pgs-gmartowski: unfortunately yes :/ I took this full configuration from Visa Checkout Integration Guide

PR #908, opened 2020-04-06
STJS-575 added services override mechanism with cybertonica mock

wirus15:


src/testing/ServicesOverrides.ts
```diff
@@ -0,0 +1,8 @@
+import { Container } from 'typedi';
+import { Cybertonica } from '../application/core/integrations/Cybertonica';
+import { CybertonicaMock } from './mocks/CybertonicaMock';
+import { environment } from '../environments/environment';
+
+if (environment.testEnvironment) {
+  Container.set(Cybertonica, CybertonicaMock);
```
wirus15, line 7: here we put all the mocks for services like Cardinal, ApplePay, VisaCheckout, etc.

PR #914, opened 2020-04-09
bugfix/STJS-609-security-code-doesnt-work-for-amex

pgs-gmartowski:


src/client/config/ConfigResolver.ts
```diff
@@ -65,9 +65,9 @@ export class ConfigResolver {
       origin: this._getValueOrDefault(config.origin, window.location.origin),
       panIcon: this._getValueOrDefault(config.panIcon, false),
       placeholders: this._getValueOrDefault(config.placeholders, {
-        pan: '',
-        expirydate: '',
-        securitycode: ''
+        pan: '***** ***** ***** *****',
```
pgs-gmartowski, line 68: @james-simpson-ST I changed default values, are they ok ?

src/application/components/security-code/SecurityCode.ts
```diff
@@ -85,7 +85,7 @@ export class SecurityCode extends FormField {
     super.setEventListener(MessageBus.EVENTS.BLUR_SECURITY_CODE);
     this._subscribeSecurityCodeChange();
     this._setDisableListener();
-    this.placeholder = this._configProvider.getConfig().placeholders.securitycode || '';
+    this.placeholder = this._configProvider.getConfig().placeholders.securitycode;
```
pgs-gmartowski, line 88: I think we don't need to check security code once again

example/json/config.json
```diff
@@ -52,14 +52,9 @@
...
-  "jwt": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJhbTAzMTAuYXV0b2FwaSIsImlhdCI6MTU4NTgzNzgwNC42ODQ3NTcyLCJwYXlsb2FkIjp7ImJhc2VhbW91bnQiOiIxMDAwIiwiYWNjb3VudHR5cGVkZXNjcmlwdGlvbiI6IkVDT00iLCJjdXJyZW5jeWlzbzNhIjoiR0JQIiwic2l0ZXJlZmVyZW5jZSI6InRlc3RfamFtZXMzODY0MSIsImxvY2FsZSI6ImVuX0dCIn19.1UMRkw6n3R2ad6MN6i9y7_BQ2M-oYANznznUgOLV5y4",
+  "jwt": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJhbTAzMTAuYXV0b2FwaSIsImlhdCI6MTU4NjQ0MjU2MC44MzQ0Mzk4LCJwYXlsb2FkIjp7ImJhc2VhbW91bnQiOiIxMDAwIiwiYWNjb3VudHR5cGVkZXNjcmlwdGlvbiI6IkVDT00iLCJjdXJyZW5jeWlzbzNhIjoiR0JQIiwic2l0ZXJlZmVyZW5jZSI6InRlc3RfamFtZXMzODY0MSIsImxvY2FsZSI6ImVuX0dCIn19.Lc8rHFAHzWA5qKCIihJYrEyXjPvtw1OgFbNk7GxyLso",
   "livestatus": 0,
   "origin": "",
-  "placeholders": {
-    "pan": "Card number",
-    "expirydate": "MM/YY",
-    "securitycode": "***"
-  },
   "panIcon": true,
```
pgs-gmartowski, line 58: I used here default values - if we want to check whole configuration then we must open ConfigSchema or ConfigResolver.

src/client/config/ConfigResolver.ts
james-simpson-ST: I think card number should be "**** **** **** ****" (4 instead of 5 in each group) as that is the most common card

james-simpson-ST:
Just need to change the card number placeholder default and then this looks good :)

src/client/config/ConfigResolver.ts
pgs-gmartowski: done, I've even moved default values to a separated file

PR #929, opened 2020-04-16
feature/STJS-609-placeholders-for-fields

pgs-gmartowski:


src/client/config/ConfigResolver.ts
```diff
@@ -50,6 +53,10 @@ export class ConfigResolver {
       translations: this._getValueOrDefault(config.translations, DefaultConfig.translations),
       visaCheckout: this._setApmConfig(config.visaCheckout, DefaultConfig.visaCheckout)
     };
+    if (!environment.production) {
+      console.error(validatedConfig);
```
pgs-gmartowski, line 57: @wirus15 - for test purposes only - we want to know what final config is

wirus15: Can we use `console.debug`?

PR #944, opened 2020-04-20
Security code checking fix

pgs-gmartowski:


tslint.json
```diff
@@ -13,7 +13,7 @@
     "no-consecutive-blank-lines": false,
     "no-console": {
       "severity": "warning",
-      "options": ["debug", "info", "log", "time", "timeEnd", "trace"]
+      "options": ["debug", "info", "time", "timeEnd", "trace"]
```
wirus15, line 16: it would be great to have this check only on commit, not during development :)

src/application/components/security-code/SecurityCode.ts
```diff
@@ -128,12 +171,15 @@ export class SecurityCode extends FormField {
   }
 
   private _subscribeSecurityCodeChange(): void {
-    this.messageBus.subscribe(MessageBus.EVENTS.CHANGE_SECURITY_CODE_LENGTH, (length: number) => {
+    this._messageBus.subscribe(MessageBus.EVENTS.CHANGE_SECURITY_CODE_LENGTH, (length: number) => {
```
wirus15, line 174: This subscriber is probably not necessary, because this class should be the only place  that emits this event. Let's just schedule the fix for the next release because there are a couple of places that needs to be checked.

src/application/components/security-code/SecurityCode.ts
```diff
@@ -27,17 +34,54 @@ export class SecurityCode extends FormField {
...
+  }
+
+  private _securityCodeUpdate$(): Observable<number> {
+    const jwtFromConfig$: Observable<string> = this._configProvider.getConfig$().pipe(map(config => config.jwt));
+    const jwtFromUpdate$: Observable<string> = this._messageBus.pipe(
+      ofType(MessageBus.EVENTS_PUBLIC.UPDATE_JWT),
+      map(event => event.data.jwt)
+    );
+    const cardNumberInput$: Observable<string> = this._messageBus.pipe(
+      tap(console.log),
```
wirus15, line 68: `console.log` => 🗑

PR #949, opened 2020-04-23
Feature/stjs 573 cardinal refactor

wirus15:


src/application/core/integrations/CardinalCommerce-copy.ts
```diff
@@ -0,0 +1,231 @@
+// import { environment } from '../../../environments/environment';
```
pgs-gmartowski, line 1: I assume that we leave this just-in-case ? 😃

src/application/core/shared/Frame.ts
```diff
@@ -32,11 +37,9 @@ export class Frame {
     new Styler(this.getAllowedStyles()).inject(this.params.styles);
   }
 
-  protected onInit() {
+  protected onInit(config?: IConfig) {
```
pgs-gmartowski, line 40: no return type specified

src/application/core/integrations/CardinalCommerce-copy.ts
wirus15: I left it "just-in-case" just for me, now it can be removed :)

wirus15:
@james-simpson-ST The unit tests are done, so the code is ready for review.

There is one test faling due to (in our opinion) invalid behavior. On the develop branch, after the cardinal validation fails, the app still performs the AUTH request. The `Merchant decline` error comes from this response. I changed the behavior so now it doesn't perform the AUTH request on failure and displays the message from cardinal validation result which is in this particular case - `Failure`.

@msakowicz the python test should be changed to check for the proper error message

james-simpson-ST:
> @james-simpson-ST The unit tests are done, so the code is ready for review.
> 
> There is one test faling due to (in our opinion) invalid behavior. On the develop branch, after the cardinal validation fails, the app still performs the AUTH request. The `Merchant decline` error comes from this response. I changed the behavior so now it doesn't perform the AUTH request on failure and displays the message from cardinal validation result which is in this particular case - `Failure`.

I think in this case we should display the message we already display for payment errors which I think is "An error occurred", the reason is I just tried it with this test card `4000000000001109` from https://cardinaldocs.atlassian.net/wiki/spaces/CCen/pages/903577725/EMV+3DS+2.0+Test+Cases and it ends up with an odd combination of "Success" and a red error message. We know that an error has occurred and the AUTH hasn't been processed so we should flag it as such

wirus15:
So in case of failed 3ds we display a generic error `An error occurred` and don't proceed with the AUTH request. Right?

src/application/core/integrations/cardinal-commerce/CardinalCommerce.ts
```diff
@@ -0,0 +1,185 @@
...
+    return this.cardinal$.pipe(
+      tap(cardinal => cardinalContinue(cardinal)),
+      tap(() => GoogleAnalytics.sendGaData('event', 'Cardinal', 'auth', 'Cardinal card authenticated')),
+      switchMap(() => this.cardinalValidated$.pipe(first())),
+      switchMap(([validationResult, jwt]: [IOnCardinalValidated, string]) => {
+        if (
+          !CardinalCommerceValidationStatus.includes(validationResult.ActionCode) ||
+          validationResult.ActionCode === 'FAILURE'
+        ) {
+          return throwError(validationResult);
```
james-simpson-ST, line 157: I believe in this case there is some missing functionality that we used to have. Basically when this errored previously if:

requestTypes = ["THREEDQUERY"] and submitOnError = true it used to redirect to the merchant with the cardinal response, now it stays on our payment form without redirecting which means that merchants performing the AUTH themselves don't have the option to decide what to do in this case

wirus15: It turned out that the fix was easy - the `COMPLETE_PAYMENT` event was not forwarded to parent frame so the submit was not being fired. Now both `submitOnError` and `submitOnSuccess` work.

james-simpson-ST:
> So in case of failed 3ds we display a generic error `An error occurred` and don't proceed with the AUTH request. Right?

Have we done this change, I think this is the last thing we need to do

wirus15:
Yes, it's right [here](https://github.com/SecureTrading/js-payments/pull/949/commits/74a2195d366765f7fc2633292dc61319dca76457#diff-04c418527d6871e669cf8ceaa0595bdcR176) :)

PR #951, opened 2020-04-23
visa-checkout-refactor

pgs-gmartowski:


pgs-gmartowski:
I've made a refactor of VisaCheckout class which is now based an abstract Apm class.
I'll try to reduce it even more.

PR #964, opened 2020-04-28
test

msakowicz:


.travis.yml
```diff
@@ -19,7 +19,8 @@ services:
 install: skip
 before_script:
   - python3 -c 'import os,sys; os.set_blocking(sys.stdout.fileno(), True)'
-  - git clone --branch=master --single-branch --depth=1 https://github.com/SecureTrading/py-payments-testing.git
+  - if [ "$TRAVIS_BRANCH" != "master" ]; then git clone --branch=test1 --single-branch --depth=1 https://github.com/SecureTrading/py-payments-testing.git; fi
```
james-simpson-ST, line 22: We also need to think about release branches. I think also they should run the master branch from py-payments-testing or do you think we should create equal release branches as we create a new release branch in the code?

msakowicz: @james-simpson-ST sorry, this is correct PR for this task https://github.com/SecureTrading/js-payments/pull/967 This branch was only for testing purpose. 
I've added additional condition also for release branches (with regex)

PR #966, opened 2020-04-28
feature/STJS-622-remove-unused-properties-in-config

pgs-gmartowski:
Let me explained what's happened here:

- `requestTypes` from main config has been removed - it's not used anywhere it's just available in API; even if merchant specifies it it's doing nothing

- `cachetoken` and `threedinit` in main config - removed - that's the same situation like in `requestTypes`;  it's available in API but was never assigned to any property / field in code, IMO useless

From my point of view we can delete those properties from config and don't need to have backward compatibility. 

Let me know guys if that make sense.

### DIFF
--- a/example/json/config.json
+++ b/example/json/config.json
@@ -23,7 +23,7 @@
   },
   "buttonId": "merchant-submit-button",
   "bypassCards": ["PIBA"],
-  "cachetoken": "",
+  "cancelCallback": null,
   "componentIds": {
     "animatedCard": "",
     "cardNumber": "",
@@ -41,6 +41,7 @@
   "datacenterurl": "",
   "deferInit": false,
   "disableNotification": false,
+  "errorCallback": null,
   "fieldsToSubmit": [
     "pan",
     "expirydate",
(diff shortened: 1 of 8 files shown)


example/json/config.json
```diff
@@ -23,7 +23,7 @@
   },
   "buttonId": "merchant-submit-button",
   "bypassCards": ["PIBA"],
-  "cachetoken": "",
+  "cancelCallback": null,
```
pgs-gmartowski, line 26: added just like `errorCallback`, `submitCallback` and `succesCallback` - to show whole config object properties

example/json/config.json
```diff
@@ -60,7 +61,6 @@
     "expirydate": "MM/YY",
     "securitycode": "***"
   },
-  "requestTypes": [],
```
pgs-gmartowski, line 63: info: can be deleted - full explanation at the top of PR

example/json/config.json
```diff
@@ -84,12 +84,12 @@
       "color-error": "#3358FF"
     }
   },
+  "submitCallback": null,
   "submitFields": [],
   "submitOnSuccess": false,
   "submitOnError": false,
   "submitOnCancel": true,
-  "submitCallback": "",
-  "threedinit": "",
```
pgs-gmartowski, line 92: info: can be deleted - full explanation at the top of PR

example/json/config.json
```diff
@@ -23,7 +23,7 @@
   },
   "buttonId": "merchant-submit-button",
   "bypassCards": ["PIBA"],
-  "cachetoken": "",
```
pgs-gmartowski, line 26: info: can be deleted - full explanation at the top of PR

PR #970, opened 2020-05-01
STJS-573 restored the deferInit functionality to make frictionless payments work

wirus15:


src/application/core/models/ISubmitData.ts
```diff
@@ -1,4 +1,5 @@
 export interface ISubmitData {
+  [index: string]: string;
```
pgs-gmartowski, line 2: it seems that rest of properties in this interface is redundant

wirus15: Not exactly. They inform us what properties we can expect in this object. The `[index: string]: string` on the other hand allows us to add more `unexpected` props :) It's not the perfect solution though, we should have specified all interfaces with fixed props for each scenario, but now is not the time to do this :)

pgs-gmartowski: true :) approved 👍

PR #982, opened 2020-05-05
STJS-632 - Updated cybertonica to have a different URL

james-simpson-ST:


src/application/core/integrations/Cybertonica.spec.ts
```diff
@@ -0,0 +1,44 @@
+import { mock } from 'ts-mockito';
+import { BrowserLocalStorage } from '../../../shared/services/storage/BrowserLocalStorage';
+import { Cybertonica } from './Cybertonica';
+
+// given
+describe('Cybertonica', () => {
+  const { instance } = cybertonicaFixture();
```
wirus15, line 7: mocks and tested instances should be created inside `beforeEach` to have a new clean version for each test

src/application/core/integrations/Cybertonica.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  // given
+  describe('_onInit', () => {
+    // when
+    beforeEach(() => {
+      // @ts-ignore
+      instance._insertCybertonicaLibrary = jest.fn();
```
wirus15, line 14: This is generally a bad approach for unit testing - it doesn't test if the class does what it suppose to, it only tests if the methods are called. We shouldn't test any of the classes privates (properties, methods). Instead we should inject dependencies like the DOM manipulation service (which is now static) and see if the calling the available public methods return expected values and makes expected calls to those dependencies.

PR #987, opened 2020-05-06
bugfix/STJS-617-bypass-cards-issue

pgs-gmartowski:


src/application/components/control-frame/ControlFrame.ts
```diff
@@ -83,23 +84,25 @@ export class ControlFrame extends Frame {
     super();
     const config$ = this._configProvider.getConfig$();
     this._communicator.whenReceive(MessageBus.EVENTS_PUBLIC.CONFIG_CHECK).thenRespond(() => config$);
-    config$.subscribe(config => this.onInit(config));
+    config$.subscribe(config => {
+      this._config = config;
```
pgs-gmartowski, line 88: info: it will be needed in couple of places in class

src/application/components/control-frame/ControlFrame.ts
```diff
@@ -83,23 +84,25 @@ export class ControlFrame extends Frame {
...
 
-  protected onInit(config: IConfig): void {
+  protected onInit(): void {
     super.onInit();
     this._setInstances();
     this._setFormFieldsValidities();
     this._formFieldChangeEvent(MessageBus.EVENTS.CHANGE_CARD_NUMBER, this._formFields.cardNumber);
     this._formFieldChangeEvent(MessageBus.EVENTS.CHANGE_EXPIRATION_DATE, this._formFields.expirationDate);
     this._formFieldChangeEvent(MessageBus.EVENTS.CHANGE_SECURITY_CODE, this._formFields.securityCode);
-    this._setRequestTypes(config);
```
pgs-gmartowski, line 96: info: it doesn't have to be called here

src/application/components/control-frame/ControlFrame.ts
```diff
@@ -127,11 +130,32 @@ export class ControlFrame extends Frame {
...
-    this._preThreeDRequestTypes = requestTypes.slice(0, threeDIndex + 1);
-    this._postThreeDRequestTypes = requestTypes.slice(threeDIndex + 1, requestTypes.length);
+  private _setPreThreeDRequestTypes(): void {
+    if (this._isCardBypassed(this._card.pan || this._getPan())) {
+      return;
+    }
+    const threeDIndex = this._config.components.requestTypes.indexOf(ControlFrame.THREEDQUERY_EVENT);
+    this._preThreeDRequestTypes = this._config.components.requestTypes.slice(0, threeDIndex + 1);
+    console.error('PRE', this._preThreeDRequestTypes);
+    console.error('POST', this._postThreeDRequestTypes);
```
pgs-gmartowski, line 140: info: temporary, for testing purposes

src/application/components/control-frame/ControlFrame.ts
```diff
@@ -127,11 +130,32 @@ export class ControlFrame extends Frame {
...
+  }
+
+  private _setPostThreeDRequestTypes(): void {
```
pgs-gmartowski, line 143: info: I've split `setRequests`; in some cases it was redundant to set both `pre` and `post` arrays

src/application/components/control-frame/ControlFrame.ts
wirus15: I belive there is no need moving this to a property and it should work just fine with arguments 😉

src/application/components/control-frame/ControlFrame.ts
wirus15: Does it make any difference at which point we set those requests? Because it seems that the only change that was necessary was to check if the card is bypassed. 

I'd go back to the original code where the config was passed as an argument. Adding properties to class that stores some values introduces state to this class and this is something we are trying to avoid as it makes testing more difficult.

james-simpson-ST:
Once the logging is removed this looks good to me 😄

PR #988, opened 2020-05-06
[WIP] Feature/stjs 625 destroy spa

wirus15:


src/application/core/shared/EventTypes.ts
```diff
@@ -0,0 +1,50 @@
...
+  SUBMIT_FORM: 'SUBMIT_FORM',
+  THREEDINIT_REQUEST: 'THREEDINIT_REQUEST',
+  THREEDINIT_RESPONSE: 'THREEDINIT_RESPONSE',
+  THREEDQUERY: 'THREEDQUERY',
+  TRANSACTION_COMPLETE: 'TRANSACTION_COMPLETE',
+  UPDATE_JWT: 'UPDATE_JWT',
+  UPDATE_MERCHANT_FIELDS: 'UPDATE_MERCHANT_FIELDS',
+  SUBSCRIBE: 'SUBSCRIBE',
+  CONFIG_CHECK: 'ST_CONFIG_CHECK'
+};
```
pgs-gmartowski, line 50: 👍

src/client/ST.ts
```diff
@@ -190,7 +189,7 @@ class ST {
     this._communicator.close();
   }
 
-  private init(): void {
+  public init(): void {
     // TODO theres probably a better way rather than having to remember to update Selectors
     Selectors.MERCHANT_FORM_SELECTOR = this._config.formId;
```
pgs-gmartowski, line 194: info: I think this is not needed anymore - now we have default values set in `ConfigResolver`

wirus15: It depends whether the app uses the value from the config or from the `Selectors`. Nevertheless this is outside the scope of this task :)

pgs-gmartowski:
👍

PR #989, opened 2020-05-08
[WIP] bugfix/STJS-634-duplicated-request-on-defer-init

pgs-gmartowski:


pgs-gmartowski:
> But I think you should revert the defer init config value to its default - `false` 😉

yes :) done

pgs-gmartowski:
closed - we've got another, better solution for this in https://github.com/SecureTrading/js-payments/pull/1012

PR #1001, opened 2020-05-11
feature/STJS-624-detect-payment-type-when-using-parenttransactionreference

pgs-gmartowski:


src/application/components/control-frame/ControlFrame.ts
```diff
@@ -361,12 +381,16 @@ export class ControlFrame extends Frame {
...
 
-  private _getPan(): string {
+  private _getPanFromJwt(): string {
     return JwtDecode<IDecodedJwt>(this.params.jwt).payload.pan
       ? JwtDecode<IDecodedJwt>(this.params.jwt).payload.pan
       : '';
   }
 
+  private _getPan(): string {
+    return this._card.pan || this._getPanFromJwt() || this._slicedPan;
```
pgs-gmartowski, line 391: info: we've got three sources of `pan` now:

- card number field
- jwt
- sliced-maskedpan from jsinit

src/application/components/control-frame/ControlFrame.ts
```diff
@@ -86,6 +87,25 @@ export class ControlFrame extends Frame {
...
+    this.messageBus
+      .pipe(
+        ofType(MessageBus.EVENTS_PUBLIC.JSINIT_RESPONSE),
+        map((event: IMessageBusEvent) => event.data.maskedpan)
+      )
+      .subscribe((maskedpan: string) => {
+        this._slicedPan = maskedpan.slice(0, 6);
+        const cvvLength: number = iinLookup.lookup(this._slicedPan).cvcLength[0];
+
+        this.messageBus.publish({
```
wirus15, line 99: this event is probably already published in `SecurityCode`

src/application/components/control-frame/ControlFrame.ts
```diff
@@ -296,7 +316,7 @@ export class ControlFrame extends Frame {
   }
 
   private _isCardWithoutCVV(): boolean {
-    const panFromJwt: string = this._getPan();
+    const panFromJwt: string = this._getPanFromJwt();
     let pan: string = '';
     if (panFromJwt || this._formFields.cardNumber.value) {
```
wirus15, line 321: I think you can simply use `getPan()` here

pgs-gmartowski: hmm, actaully I've got similar function in this class:
`private _getPan(): string {
    return this._card.pan || this._getPanFromJwt() || this._slicedPan;
  }`
should I still add getPan() instead of _getPanFromJwt ?

src/application/components/control-frame/ControlFrame.ts
pgs-gmartowski: true, done

PR #1008, opened 2020-05-13
bugfix/STJS-648-submitoncancel-doesnt-work-on-apple-pay

pgs-gmartowski:


src/client/classes/CommonFrames.class.ts
```diff
@@ -140,19 +140,39 @@ export class CommonFrames extends RegisterFrames {
...
       if (this._submitOnCancel) {
-        formData = Object.assign(data, { errormessage: Language.translations.PAYMENT_CANCELLED });
+        this._submitForm(data);
       }
-      DomMethods.addDataToForm(form, formData, this._getSubmitFields(data));
-      form.submit();
+      return;
     }
+
+    if (data.errorcode !== '0' && data.errorcode !== 'cancelled') {
```
wirus15, line 164: you don't need to check if `data.errorcode !== 'cancelled'` because it's already checked earlier

pgs-gmartowski: 🤦‍♂️ true, fixed

PR #1037, opened 2020-05-18
STJS-617 fixed diving pre and post request types

wirus15:


src/application/components/control-frame/ControlFrame.ts
```diff
@@ -158,39 +157,22 @@ export class ControlFrame extends Frame {
       ControlFrame._resetJwt();
     });
   }
+
   private _setRequestTypes(config: IConfig): void {
-    const requestTypes = config.components.requestTypes;
+    const skipThreeDQuery = this._isCardBypassed(this._getPan());
+    const filterThreeDQuery = (requestType: string) =>
+      !skipThreeDQuery || requestType !== ControlFrame.THREEDQUERY_EVENT;
+    const requestTypes = [...config.components.requestTypes].filter(filterThreeDQuery);
```
pgs-gmartowski, line 165: not sure if it wouldn't be redundant but there are no types in each `const`

wirus15: They are not necessary as long as the compiler can resolve them by itself. They make the code more readable, however. I always put types in function declarations and inside the function's body - not always :)

PR #1038, opened 2020-05-19
fix-packages-vulns

pgs-gmartowski:


pgs-gmartowski:
@wirus15 there is high vulnerability in one of packages (wbpack-dev-server), I've fixed it

PR #1044, opened 2020-05-22
STJS-653 - Just processing THREEDQUERY does not work

james-simpson-ST:


example/index.html
```diff
@@ -11,7 +11,7 @@
 </head>
 
 <body>
-<form id="st-form" class="example-form" autocomplete="off" novalidate>
+<form id="st-form" class="example-form" autocomplete="off" action="https://example.com" novalidate>
```
james-simpson-ST, line 14: Need to revert but used during testing to avoid infinite redirect loop in test

example/json/config.json
```diff
@@ -103,8 +93,8 @@
   },
   "submitCallback": null,
   "submitFields": [],
-  "submitOnSuccess": false,
-  "submitOnError": false,
+  "submitOnSuccess": true,
```
james-simpson-ST, line 96: Need to revert but used during testing

src/application/components/control-frame/ControlFrame.ts
```diff
@@ -272,26 +272,53 @@ export class ControlFrame extends Frame {
...
+        .catch((error: any) => {
+          this.messageBus.publish({ type: MessageBus.EVENTS_PUBLIC.CALL_MERCHANT_ERROR_CALLBACK }, true);
+          this._notification.error(Language.translations.PAYMENT_ERROR);
+          this._validation.blockForm(FormState.AVAILABLE);
+        })
+        .finally(() => {
+          ControlFrame._resetJwt();
+        });
+    } else {
+      //  TODO is this matching the previous behaviour exactly, we need to compare for submitOnSuccess true&false and submitOnError true&false
```
james-simpson-ST, line 297: This doesn't feel like a nice solution but can't see a better way at the moment

PR #1060, opened 2020-05-28
feature/STJS-578-implement-lgtm-suggestions

pgs-gmartowski:


pgs-gmartowski:
@wirus15 , @james-simpson-ST I think that's all on javascript side; the second part is python repo

PR #1076, opened 2020-06-01
STJS-636 added cybertonica timeout

wirus15:
The payment was freezing when the cybertonica script could not be loaded (eg. blocked by the ad-block). I added a simple timeout so that when the TID is not resolved within 5 seconds it just skips the cybertonica completely and sends the payment without TID.

### DIFF
--- a/src/application/core/integrations/Cybertonica.ts
+++ b/src/application/core/integrations/Cybertonica.ts
@@ -42,7 +42,6 @@ export class Cybertonica implements ICybertonica {
 
     this.tid = Promise.race([tid, timeout]);
     this.tid.then(value => this.storage.setItem(Cybertonica.TID_KEY, value));
-    this.tid.then(value => console.log('TID', value));
 
     return this.tid;
   }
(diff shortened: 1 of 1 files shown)


src/application/core/integrations/Cybertonica.ts
```diff
@@ -35,24 +35,24 @@ export class Cybertonica implements ICybertonica {
...
   public init(apiUserName: string): Promise<string> {
-    this.tid = this._insertCybertonicaLibrary().then(
-      () => AFCYBERTONICA.init(apiUserName, undefined, Cybertonica.getBasename()) || this.initFailed()
+    const tid = this._insertCybertonicaLibrary().then(() =>
+      AFCYBERTONICA.init(apiUserName, undefined, Cybertonica.getBasename())
     );
-    this.tid.then(tid => this.storage.setItem(Cybertonica.TID_KEY, tid));
+    const timeout = new Promise(resolve => setTimeout(() => resolve(null), Cybertonica.TID_TIMEOUT));
+
+    this.tid = Promise.race([tid, timeout]);
```
pgs-gmartowski, line 43: 👍

PR #1081, opened 2020-06-03
feature/STJS-633-national-zakat-foundation-block-form

pgs-gmartowski:


src/client/classes/CardFrames.class.ts
```diff
@@ -53,6 +54,7 @@ export class CardFrames extends RegisterFrames {
   private _loadAnimatedCard: boolean;
 
   constructor(
+    private _sessionStorage: BrowserSessionStorage,
```
wirus15, line 57: Why did you add this?

pgs-gmartowski: 🤣 deleted

pgs-gmartowski:
> I was wondering if this could be done sooner and it looks like I was right. If you look at the `ST.Components(...)` method you will see that the `CardFrames.init(...)` is being called after the control frame is loaded and synchronized with config. This is fine if the submit button is generated by our library, but in most cases the button is provided by merchants. You could try to disable the button somewhere in the `Components(...)` call without waiting for control frame.

I checked this and button is blocked slightly faster now.

src/client/ST.ts
```diff
@@ -335,6 +338,25 @@ class ST {
...
   }
+
+  private blockSubmitButton() {
+    const form: HTMLFormElement = document.getElementById(this._config.formId) as HTMLFormElement;
+
+    if (!form) {
+      return;
+    }
+
+    let submitButton: HTMLInputElement | HTMLButtonElement = this._config.buttonId
```
wirus15, line 349: could be shorter:
```typescript
const submitButton: HTMLInputElement | HTMLButtonElement = document.getElementById(this._config.buttonId) ||
  form.querySelector(ST.BUTTON_SUBMIT_SELECTOR) || 
  form.querySelector(ST.INPUT_SUBMIT_SELECTOR);

pgs-gmartowski: true :) done

PR #1082, opened 2020-06-03
[WIP] add-config.json-file-to-gitignore

pgs-gmartowski:


pgs-gmartowski:
> We should leave a `config.json.dist` for example without jwt as a template for the `config.json`

True, added.

example/config.json.dist
```diff
@@ -0,0 +1,133 @@
...
+            "supportedNetworks": [],
+            "total": {
+                "label": "Secure Trading Merchant",
+                "amount": "10.00"
+            }
+        },
+        "placement": "st-apple-pay"
+    },
+    "buttonId": "merchant-submit-button",
+    "bypassCards": [
```
wirus15, line 30: these cards should not be bypassed by default - I think only "PIBA" should be here

example/config.json.dist
```diff
@@ -0,0 +1,133 @@
...
+        "animatedCard": "",
+        "cardNumber": "",
+        "expirationDate": "",
+        "notificationFrame": "",
+        "securityCode": ""
+    },
+    "components": {
+        "defaultPaymentType": "",
+        "requestTypes": [
+            "THREEDQUERY",
```
wirus15, line 49: I believe the correct order should be "RISKDEC","ACCOUNTCHECK","THREEDQUERY","AUTH"

example/config.json.dist
pgs-gmartowski: true, deleted

example/config.json.dist
pgs-gmartowski: yes, or as I remember `RISKDEC` can be at the and (but not sure if with the `AUTH` in array).
Fixed.

PR #1087, opened 2020-06-05
STJS-586 sentry

wirus15:


src/client/config/ConfigResolver.ts
```diff
@@ -50,7 +50,8 @@ export class ConfigResolver {
       submitOnSuccess: this._getValueOrDefault(config.submitOnSuccess, DefaultConfig.submitOnSuccess),
       successCallback: this._getValueOrDefault(config.successCallback, DefaultConfig.successCallback),
       translations: this._getValueOrDefault(config.translations, DefaultConfig.translations),
-      visaCheckout: this._setVisaCheckoutConfig(config.visaCheckout, DefaultConfig.visaCheckout)
+      visaCheckout: this._setVisaCheckoutConfig(config.visaCheckout, DefaultConfig.visaCheckout),
+      sentryDsn: this._getValueOrDefault(config.sentryDsn, '')
```
pgs-gmartowski, line 54: I think this should be in alphabetical order, if we had such convention before

pgs-gmartowski: let's put default value of `sentryDsn` into `DefaultConfig` as well

src/shared/services/sentry/EventScrubber.spec.ts
```diff
@@ -0,0 +1,42 @@
+import { Event } from '@sentry/types';
+import { EventScrubber } from './EventScrubber';
+
+describe('EventScrubber', () => {
```
pgs-gmartowski, line 4: I was using `//given // when // then` convention before; should we continue this or leave it ?

src/shared/services/sentry/EventScrubber.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+        config: {
+          jwt: 'some-long-jwt-value',
+          foo: 'bar'
+        }
+      }
+    };
+
+    const result = eventScrubber.scrub(event);
+
+    expect(result.extra.config).toEqual({
```
pgs-gmartowski, line 23: you're using only `config` property here, so I would do sth like:
`
    const { config }= eventScrubber.scrub(event).extra;

    expect(config).toEqual({
`
but it's nth :)

src/shared/services/sentry/EventScrubber.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    const event: Event = {
+      request: {
+        url: 'https://webservices.securetrading.net?foo=bar&jwt=some-long-jwt&xyz=abc',
+        query_string: 'jwt=some-long-jwt'
+      }
+    };
+
+    const result = eventScrubber.scrub(event);
+
+    expect(result.request.url).toBe('https://webservices.securetrading.net?foo=bar&jwt=*****&xyz=abc');
```
pgs-gmartowski, line 39: looks like `https://webservices.securetrading.net?foo=bar` is repeated twice; maybe it's good to use template strings here

wirus15:
@pgs-gmartowski I made changes from your review. Plus I moved the sentry dsn to environment and added `errorReporting` property to config, like @james-simpson-ST suggested.

example/json/config.json
```diff
@@ -59,6 +59,7 @@
   "deferInit": false,
   "disableNotification": false,
   "errorCallback": null,
+  "errorReporting": false,
```
pgs-gmartowski, line 62: I think James suggested that it should be named `debug`

src/shared/services/sentry/SentryService.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    this.sentry.setTag('hostName', this.sentryContext.getHostName());
+    this.sentry.setTag('frameName', this.sentryContext.getFrameName());
+
+    this.sentry.init({
+      dsn,
+      environment: this.sentryContext.getEnvironmentName(),
+      release: this.sentryContext.getReleaseVersion(),
+      whitelistUrls: ['https://webservices.securetrading.net'],
```
pgs-gmartowski, line 29: just wondering if such constant values should be somewhere outside of class, or inside as static properties ?
What should be our approach ?

wirus15: good point 👍

example/json/config.json
wirus15: I remember he suggested `debug` but I thought `errorReporting` better describes what this property does. We could use `debug` to put more verbose `console.logs`.

PR #1100, opened 2020-06-09
[WIP] bugfix/STJS-673-notification-banner

pgs-gmartowski:


pgs-gmartowski:
> You should probably remove the autohide timeout when success message appears

True, fixed.

PR #1147, opened 2020-07-01
[WIP] Feature/stjs 482 redux

wirus15:
implemented features:
- `Store` - based on redux service to hold and manage the application state. This should be the only service that has a state, all other services should be stateless. The state is shared across all iframes but not with the parent frame (we want to reduce the role of the parent frame in the next stages).
- `state, actions, reducers` - according to the redux pattern, the state should only be modified by the reducers in response to actions. The state is strongly typed. I also implemented a strong typing for actions, which is something I wanted to do with the MessageBus events (the action payload's type is being checked based on the action type).
- `rxjs integration` - added extra features to the store that utilises rxjs observables. The whole store can be subscribed as well as a part of it using the `.select$(...)` method.
- `redux-observable` - an extension to redux that allows implementation of side effects. Side effect is a situation when one action causes a dispatch of another one
- `redux-logger` - another extension for debugging purposes - it logs all actions and state changes caused by them to the browser's console.


james-simpson-ST:
I think I am not understanding something with the implementation, I can see that we check that we are the control frame in StoreAccessor which I guess is how we are preventing the parent from from modifying or viewing the state stored within the application logic?

However I can't see how we still allow the frames like pan etc to still use the state throughout the app as they won't be considered a control frames unless I am missing something?

Are we intending to use this to pass the value of the pan/securitycode etc throughout the ST hosted frames instead of the message bus, in which case I guess these other frames will need to be able to modify the state when the user changes the pan and the animated card frame will need to be able to view the state when trying to display the updated card details

But we definitely need to ensure the parent frame will not be able to access this state as this is what makes our library PCI SAQ A because the merchants page doesn't have access to the card details

wirus15:
The `StoreAccessor` creates an instance of the redux store and places it in the control frame's window object. Our `Store` service is a wrapper on the redux's one which accesses it through the store accessor. This is only possible for the component frames as they are in the same domain as the control frame. The parent frame won't be able to access the internal redux store.

Each component frame has their own `Store` instance, but they all forward the requests to the single redux store inside the control frame's window object. This applies to both reading (getState(), subscribe(), select()) and writing (by dispatching actions). An action dispatched in the CardNumber frame is being forwarded to the redux store in the control frame. A proper reducer then modifies the state. Finally all frames that subscribed to the changes of the entire state (or even a single property) is being notified about that change and can react accordingly.

The difference from the MessageBus is that both events (actions) the state (like a pan number) is now shared across all frames. In message bus only the events were shared, but each frame could have their own different state based on those events. It caused problems such as missing the event being fired before the frame loaded and started listening for that event. Now the frame can access the state (like config or jwt) anytime and get notified whenever this state changes.

src/application/core/store/epics/dummyEpic.ts
```diff
@@ -0,0 +1,13 @@
...
+import { ActionName, IAction } from '../IAction';
+import { UPDATE_CONFIG } from '../reducers/config/ConfigActions';
+import { ofType } from '../operators/ofType';
+import { EMPTY } from 'rxjs';
+
+export const dummyEpic: Epic<IAction> = action$ =>
+  action$.pipe(
+    ofType(UPDATE_CONFIG),
+    mapTo({ type: 'FOO' as ActionName })
+    // switchMap(() => EMPTY),
```
pgs-gmartowski, line 12: should this still be commented out ?

wirus15: This is just a dummy to test if it works. It will be removed

PR #1150, opened 2020-07-02
bugfix/STJS-691-updatejwt-to-set-amex-card-as-parent-does-not-update-securitycode-field

pgs-gmartowski:


src/application/components/security-code/SecurityCode.ts
```diff
@@ -83,16 +86,16 @@ export class SecurityCode extends FormField {
       map(jwt => JwtDecode<IDecodedJwt>(jwt).payload.pan)
     );
 
-    const maskedPanFromJsInit$: Observable<string> = this._sessionStorage.select(store => store['app.maskedpan']);
+    // const maskedPanFromJsInit$: Observable<string> = this._sessionStorage.select(store => store['app.maskedpan']);
```
pgs-gmartowski, line 89: info: this shouldn't be needed anymore

PR #1162, opened 2020-07-06
synchronize-develop-with-master

pgs-gmartowski:


pgs-gmartowski:
Needed to be synchronized with master after finishing - https://github.com/SecureTrading/js-payments/pull/1161

PR #1163, opened 2020-07-06
Feature/stjs 482 redux config

wirus15:


src/application/components/control-frame/ControlFrame.ts
```diff
@@ -105,6 +107,14 @@ export class ControlFrame extends Frame {
         });
       });
 
+    this.messageBus.pipe(ofType(PUBLIC_EVENTS.CONFIG_CHANGED)).subscribe((event: IMessageBusEvent<IConfig>) => {
+      if (event.data) {
+        this._store.dispatch({ type: UPDATE_CONFIG, payload: event.data });
+      } else {
+        this._store.dispatch({ type: CLEAR_CONFIG });
+      }
```
pgs-gmartowski, line 115: or:
```
  if (event.data) {
        this._store.dispatch({ type: UPDATE_CONFIG, payload: event.data });
        return;
      }
      this._store.dispatch({ type: CLEAR_CONFIG });
      
```

src/application/components/control-frame/ControlFrame.ts
```diff
@@ -105,6 +107,14 @@ export class ControlFrame extends Frame {
         });
       });
 
+    this.messageBus.pipe(ofType(PUBLIC_EVENTS.CONFIG_CHANGED)).subscribe((event: IMessageBusEvent<IConfig>) => {
```
pgs-gmartowski, line 110: nice to have - we can make destructuring here if we don't use whole event object:
`.subscribe(({data}) => {`

wirus15: I'd stick to the original version. With destructuring there is a little less code, but we loose the type control and some readability.

PR #1168, opened 2020-07-09
feature/STJS-675-improve-elements-class

pgs-gmartowski:


src/client/classes/element/Element.ts
```diff
@@ -0,0 +1,44 @@
+import { IStyle } from '../../../shared/model/config/IStyle';
+import { Selectors } from '../../../application/core/shared/Selectors';
+import { IElementAttributes } from './IElementAttributes';
+
+export class Element {
+  private readonly _attributes: IElementAttributes;
+  private readonly _src: string;
+
+  constructor(name: string, id: string, styles?: IStyle, params?: {}, tabindex?: string) {
+    const urls = new Map(
```
wirus15, line 10: could be a static property

src/client/classes/element/Element.ts
```diff
@@ -0,0 +1,44 @@
+import { IStyle } from '../../../shared/model/config/IStyle';
+import { Selectors } from '../../../application/core/shared/Selectors';
+import { IElementAttributes } from './IElementAttributes';
+
+export class Element {
```
wirus15, line 5: This class is more like an iframe factory and that's how I'd name it. Plus it doesn't need a stateful instance for each iframe. It should look like this:

```typescript
@Service
class IframeFactory {
  create(name: string, id: string, styles?: IStyle, params?: {}, tabindex?: string): HTMLIFrameElement { 
    ...
  }
}

File: src/client/classes/element/Element.ts [hunk 3]
```diff
@@ -0,0 +1,44 @@
...
+  private readonly _attributes: IElementAttributes;
+  private readonly _src: string;
+
+  constructor(name: string, id: string, styles?: IStyle, params?: {}, tabindex?: string) {
```
wirus15, line 9: tabindex imo should be a `number`. And it's name should be in `camelCase`

src/client/classes/element/IElementAttributes.ts
```diff
@@ -0,0 +1,10 @@
+export interface IElementAttributes {
```
wirus15, line 1: I think that objects we use in the code should keep the notation. That means the names should be in `camelCase` and values should have proper types (bool, number). Then you map them to html attributes.

pgs-gmartowski: done

src/client/classes/element/Element.ts
pgs-gmartowski: true, also changed `allowTransparency `and `frameBorder`

src/client/classes/element/Element.ts
pgs-gmartowski: true, added and changed

src/client/classes/element/Element.ts
pgs-gmartowski: fixed

src/application/core/shared/frame/Frame.spec.ts
```diff
@@ -22,7 +22,7 @@ describe('Frame', () => {
         }
       }
     ]
-  ]).test('Frame.parseUrl', (url, expected) => {
+  ]).test('frame.parseUrl', (url, expected) => {
```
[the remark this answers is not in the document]
pgs-gmartowski: yes, I suppose it was refactored automatically

src/client/classes/card-frames/CardFrames.class.ts
```diff
@@ -184,32 +184,22 @@ export class CardFrames extends RegisterFrames {
...
-    );
+    this._cardNumber = this._iframeFactory.create('cardNumber', 'st-card-number-iframe', styles, this.params);
     this.elementsToRegister.push(this._cardNumber);
   }
 
   private _initExpiryDateFrame(styles: {}): void {
     this._expirationDate = this._iframeFactory.create(
-      Selectors.EXPIRATION_DATE_COMPONENT_NAME,
-      Selectors.EXPIRATION_DATE_IFRAME,
+      'expirationDate',
```
wirus15, line 193: why you no longer use constants from the `Selectors` class?

src/client/ST.ts
```diff
@@ -288,7 +292,7 @@ class ST {
       /* tslint:disable:no-console */
       console.log(
         '%cThe %csecure%c//%ctrading %cLibrary is currently working in test mode. Please check your configuration.',
-        'margin: 100px 0; font-size: 2em; color: #e71b5a',
+        'margin: 100px 0; font-size: 2em; colQ  or: #e71b5a',
```
wirus15, line 295: `colQ or`? 😁

src/client/classes/element/IframeFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+    })
+  );
+
+  create(name: string, id: string, styles?: IStyle, params?: {}, tabIndex?: number): HTMLIFrameElement {
+    const componentParams = new URLSearchParams(params).toString();
+    const componentStyles = new URLSearchParams(styles).toString();
+    const componentAddress = IframeFactory.URLS.get(name);
+    const iframe = document.createElement('iframe');
+
+    IframeFactory.SRC = `${componentAddress}?${componentStyles}${componentParams ? '&' + componentParams : ''}`;
```
wirus15, line 26: why did you put everything as static properties? modified every time an iframe is created? this makes absolutely no sense... 😞

src/client/classes/element/IframeFactory.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+
+// given
+describe('IframeFactory', () => {
+  let instance: IframeFactory;
+  let iframe: HTMLIFrameElement;
+
+  // when
+  beforeEach(() => {
+    instance = new IframeFactory();
+    iframe = instance.create(
```
wirus15, line 11: the iframe should be created inside the test (`it(...)`), not only checked

src/client/classes/card-frames/CardFrames.spec.ts
```diff
@@ -1,20 +1,23 @@
...
+import { anyString, instance as instanceOf, mock, when } from 'ts-mockito';
 import { of } from 'rxjs';
+import { IframeFactory } from '../element/IframeFactory';
 
-jest.mock('./../../../src/application/core/shared/Notification/Notification');
+jest.mock('./../../../../src/application/core/shared/notification/Notification');
 
 // given
 describe('CardFrames', () => {
+  const { instance } = cardFramesFixture();
```
wirus15, line 16: instance should be created and initialized in beforeEach

src/client/classes/card-frames/CardFrames.spec.ts
```diff
@@ -365,6 +359,40 @@ function cardFramesFixture() {
       placeholders: { pan: 'Card number', expirydate: 'MM/YY', securitycode: '***' }
     })
   );
+  when(iframeFactory.create(anyString(), anyString())).thenCall((name: string, id: string) => {
+    const iframe: HTMLIFrameElement = document.createElement('iframe');
+    iframe.setAttribute('name', name);
+    iframe.setAttribute('id', id);
+    return iframe;
+  });
+  iframeFactory.create(
```
wirus15, line 368: why do you create iframes inside the test? this should be the job of the class under test

src/client/ST.ts
pgs-gmartowski: 🤦‍♂️ fixed

src/client/classes/card-frames/CardFrames.class.ts
pgs-gmartowski: I think this was my unit tests debugging changes, fixed

src/client/classes/element/IframeFactory.ts
pgs-gmartowski: changed

src/client/classes/element/IframeFactory.spec.ts
pgs-gmartowski: fixed

src/client/classes/card-frames/CardFrames.spec.ts
pgs-gmartowski: fixed

src/client/classes/card-frames/CardFrames.spec.ts
pgs-gmartowski: fixed

src/application/core/shared/input/Input.ts
```diff
@@ -77,15 +83,36 @@ export class FormField extends Frame {
...
       'font-size-message': { property: 'font-size', selector: message },
+      'font-family-input': { property: 'font-family', selector: input },
+      'font-family-input-error': { property: 'font-family', selector: inputError },
+      'font-family-label': { property: 'font-family', selector: label },
+      'font-family-message': { property: 'font-family', selector: message },
       'line-height-input': { property: 'line-height', selector: input },
       'line-height-input-error': { property: 'line-height', selector: inputError },
       'line-height-label': { property: 'line-height', selector: label },
       'line-height-message': { property: 'line-height', selector: message },
+      'margin-input': { property: 'margin', selector: input },
```
james-simpson-ST, line 94: Isn't this already an option as space-outset-input? Maybe we should keep all of the margin ones the same so we should add:

space-inset-message
space-outset-message

And remove margin-input (as its the same as space-outset-input)

src/application/core/shared/notification/Notification.ts
```diff
@@ -39,8 +42,11 @@ export class Notification {
 
   private _applyStyles(): void {
     this._configProvider.getConfig$().subscribe((config: IConfig) => {
-      // @ts-ignore
-      new Styler(this._getAllowedStyles()).inject(config.styles.notificationFrame);
+      console.error(config.styles.notificationFrame);
```
james-simpson-ST, line 45: Should go through and make sure we remove any consoles we don't need anymore, can't see why we want this one

src/shared/services/message-bus/FramesHub.ts
```diff
@@ -55,13 +55,12 @@ export class FramesHub {
...
   public notifyReadyState(): void {
     const frameName = this.identifier.getFrameName();
-
     if (frameName === Selectors.MERCHANT_PARENT_FRAME) {
       return;
     }
 
     if (!frameName) {
-      // @todo: Validation class should not extend Frame class. Once fixed this line should throw an error */
+      // @todo: Validation class should not extend frame class. Once fixed this line should throw an error */
```
james-simpson-ST, line 63: I believe you've stoped this from extending frame, can this @todo be done now?

src/application/components/card-number/CardNumber.ts
```diff
@@ -57,6 +63,17 @@ export class CardNumber extends FormField {
     );
     this._sendState();
     this._inputElement.setAttribute(CardNumber.PLACEHOLDER_ATTRIBUTE, this.placeholder);
+    this.configProvider.getConfig$().subscribe((config: IConfig) => {
+      const styler: Styler = new Styler(this.getAllowedStyles(), this.frame.parseUrl().styles);
+      if (styler.isVertical(config.styles.cardNumber)) {
```
james-simpson-ST, line 68: I'm not sure I understand what this isVertical is for, can you explain what it is for and when it is true/false?

src/shared/services/message-bus/FramesHub.ts
pgs-gmartowski: True :) 
I deleted this `if`, hope this is a last class without name which extended `Frame` class

src/application/core/shared/notification/Notification.ts
pgs-gmartowski: :/ I must forgot about this one (I always search for console.errors/logs before committing changes).
Fixed.

src/application/core/shared/input/Input.ts
pgs-gmartowski: true, I didn't see that there is another property for margin and padding.
Fixed.

src/application/components/card-number/CardNumber.ts
pgs-gmartowski: ah, should be horizontal of course :) maybe we can call this `isDirectionRow` or maybe better `isLinedUp` ?
I also refactor this function because it repeats in three different classes; it's now moved to Styler.ts and called `lineUp`.

pgs-gmartowski:
> Code looks fine to me, but the pipeline seems to be failing on chrome tests, is this something Marek has to fix?

Good question; those two cases which are failing looks like a new features. I ask Marek about it.

PR #1181, opened 2020-07-14
moved storage from browser to redux

wirus15:


src/application/core/shared/Notification.ts
```diff
@@ -32,13 +34,13 @@ export class Notification {
...
       this._translator = new Translator(this._browserLocalStorage.getItem('locale'));
     });
-
-    this._applyStyles();
   }
 
   private _applyStyles(): void {
-    // @ts-ignore
-    new Styler(this._getAllowedStyles()).inject(this._configProvider.getConfig().styles.notificationFrame);
+    this._configProvider.getConfig$().subscribe(config => {
```
pgs-gmartowski, line 40: we can use destructuring here, or add `IConfig`

PR #1183, opened 2020-07-15
added additional popup for testing purpose

msakowicz:


example/index.html
```diff
@@ -142,6 +142,8 @@ <h1 class="example-form__title">
         div.style.backgroundColor = tp;
         if (tp === 'green') {
```
pgs-gmartowski, line 143: you can use `switch - case` as well

example/index.html
```diff
@@ -157,6 +159,7 @@ <h1 class="example-form__title">
       stConfig.submitCallback = function someFancyfunction(data) {
         var stringified = JSON.stringify(data);
         var testVariable = 'This is what we have got after submit' + stringified;
+        displayPopup('data-popup', 'Error code: ' + (data.errorcode == "0" ? "OK" : "Error"), 'blue');
```
wirus15, line 162: you could write 
```typescript
(data && data.errorcode == '0' ? 'OK' : 'Error')
```
to avoid errors when `data` is not set

PR #1247, opened 2020-08-10
 Feature/stjs 855 convert classes into const files [Selectors.ts, EventTypes.ts and Language.ts]

pgs-gmartowski:


src/application/core/models/constants/Selectors.ts
```diff
@@ -1,48 +1,85 @@
 // @TODO: changed implementation and references to const
```
james-simpson-ST, line 1: Isn't this TODO done now?

pgs-gmartowski: good point ! :) I forgot to delete this. 
Removed.

